PR #117, opened 2018-04-23
Add `preferredPosition` to polaris-popover

tomnez:
## Overview

This adds the `preferredPosition` to the [polaris-popover component](https://polaris.shopify.com/components/overlays/popover).

It basically translates the user-provided `preferredPosition` value into a `verticalPosition` attribute since we use [ember-basic-dropdown](http://ember-basic-dropdown.com/) as our popover component under the hood.

Test for this is skipped because ember-basic-dropdown uses ember wormhole for the popover content, which renders content outside of the ember testing container.

### DIFF
--- a/addon/components/polaris-popover.js
+++ b/addon/components/polaris-popover.js
@@ -6,6 +6,8 @@ import { warn } from '@ember/debug';
 import layout from '../templates/components/polaris-popover';
 
 const { ViewUtils } = Ember;
+const ABOVE = 'above';
+const BELOW = 'below';
 
 /**
  * Polaris popover component.
@@ -39,9 +41,7 @@ export default Component.extend({
    * @type {string}
    * @default 'below'
    */
-  preferredPosition: computed(function() {
-    return 'below';
-  }),
+  preferredPosition: BELOW,
 
   /**
    * Show or hide the Popover
@@ -95,17 +95,29 @@ export default Component.extend({
   /*
    * Internal properties.
    */
-  verticalPosition: computed('preferredPosition', function() {
-    let preferredPosition = this.get('preferredPosition');
+  verticalPosition: computed('preferredPosition', {
+    // If `preferredPosition` is set to `mostSpace`, the value
+    // will be calculated and set when the user opens the popover.
+    // The only allowed values are 'above' and 'below', so we
+    // return null for anything other than those values and let
+    // ember-basic-dropdown use its default value.
+    get() {
+      let preferredPosition = this.get('preferredPosition');
 
-    if (preferredPosition === 'above' || preferredPosition === 'below') {
-      return preferredPosition;
+      if (preferredPosition === ABOVE || preferredPosition === BELOW) {
+        return preferredPosition;
+      }
+
+      return null;
+    },
+
+    set(key, value) {
+      if (value === ABOVE || value === BELOW) {
+        return value;
+      }
+
+      return null;
     }
-
-    // If set to `mostSpace`, the value will be calculated
-    // and set when the user opens the popover.
-
-    return null;
   }),
 
   triggerStyle: computed(function() {
@@ -137,7 +149,7 @@ export default Component.extend({
 
     if (activators.length > 1) {
       warn('Multiple popover activators found. Defaulting to `preferredPosition` of `below`');
-      return 'below';
+      return BELOW;
     }
 
     let [activator] = activators;
@@ -146,7 +158,7 @@ export default Component.extend({
     let bottomSpace = windowHeight - bottom;
 
     // Use `below` if distance is equal
-    return bottomSpace >= top ? 'below' : 'above';
+    return bottomSpace >= top ? BELOW : ABOVE;
   },
 
   actions: {
--- a/addon/templates/components/polaris-popover.hbs
+++ b/addon/templates/components/polaris-popover.hbs
@@ -1,6 +1,6 @@
 {{#basic-dropdown
   verticalPosition=verticalPosition
-  onOpen=(action 'onOpen')
+  onOpen=(action "onOpen")
   as |dd|
 }}
   {{yield (hash


addon/components/polaris-popover.js
```diff
@@ -90,11 +93,73 @@ export default Component.extend({
...
+
+    if (preferredPosition === 'below') {
+      return preferredPosition;
+    }
+
+    // If set to `mostSpace`, the value will be calculated
+    // and set when the user opens the popover.
+
+    return null;
+  }),
```
andrewpye: `.readOnly()`?

addon/components/polaris-popover.js
```diff
@@ -90,11 +93,73 @@ export default Component.extend({
...
+  verticalPosition: computed('preferredPosition', function() {
+    let preferredPosition = this.get('preferredPosition');
+
+    if (preferredPosition === 'above') {
+      return preferredPosition;
+    }
+
+    if (preferredPosition === 'below') {
+      return preferredPosition;
+    }
```
andrewpye: Why not bundle these two into a single `if`-statement?

addon/components/polaris-popover.js
```diff
@@ -90,11 +93,73 @@ export default Component.extend({
...
+  /**
+   * Checks the dropdown trigger's location on
+   * screen to determine which vertical direction
+   * has more space to open the dropdown.
+   */
+  mostVerticalSpace() {
+    let component;
+
+    // Hack to get the component's container
+    // element since this component uses `tagName: ''`
```
andrewpye: Add a link here to the RFC I linked you to so we know where this came from in the future 👍

addon/components/polaris-popover.js
```diff
@@ -90,11 +93,73 @@ export default Component.extend({
...
+
+  actions: {
+    onOpen() {
+      // Check to see if `preferredPosition` is set to `mostSpace`
+      // onOpen, since user could have scrolled vertically since
+      // the time the component originally rendered.
+      let preferredPosition = this.get('preferredPosition');
+
+      if (preferredPosition === 'mostSpace') {
+        this.set('verticalPosition', this.mostVerticalSpace());
```
andrewpye: Not a huge fan of the naming of `mostVerticalSpace` as a method - it sounds more like a property since there's no verb in the name. How about `getMostVerticalSpace` or making it a `volatile` CP and keeping the `mostVerticalSpace` name?

docs/popover.md
```diff
@@ -6,7 +6,7 @@
 component](https://polaris.shopify.com/components/overlays/popover). This component uses [`ember-basic-dropdown`](https://github.com/cibernox/ember-basic-dropdown) to implement popover functionality. Note that the usage is slightly different from the React implementation so please pay attention to the examples below.
 
 
-**NOTE:** _the `preferredPosition`, `active`, `activatorWrapper`, `preventAutofocus` and `onClose` properties are currently unimplemented._
+**NOTE:** _the `active`, `activatorWrapper`, `preventAutofocus` and `onClose` properties are currently unimplemented._
```
andrewpye: 👍

tests/integration/components/polaris-popover-test.js
```diff
@@ -101,3 +103,29 @@ test('it renders the correct HTML with sectioned attribute', function(assert) {
   assert.equal(popoverSections.length, 1, 'renders one popover section');
   assert.equal(popoverSections[0].textContent.trim(), 'This is some sectioned popover content', 'popover section contains the correct content');
 });
+
+skip('it renders the correct HTML with `prefferedPosition` attribute', function(assert) {
```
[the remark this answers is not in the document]
vladucu: yes, I'd lean towards using `document.querySelector` (or whatever works) and not skip the test if possible

addon/components/polaris-popover.js
vladucu: good with me for now, ideally I think we should investigate refactoring this at some point and not using `ember-basic-dropdown`

addon/components/polaris-popover.js
tomnez: This gets `set` by the `onOpen` action if the position is set to `"mostSpace"`, so it can't be readOnly

tests/integration/components/polaris-popover-test.js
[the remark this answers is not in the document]
tomnez: 👍 I'll see if `document.querySelector` works

addon/components/polaris-popover.js
andrewpye: Hmm... seems like an _unlikely_ scenario, but `set`ting this in `onOpen` will mean that if the host app passes in a dynamic `preferredPosition` and that value changes from 'mostSpace' after first open, the popover won't honour that change because `onOpen` will overwrite the CP behaviour here. You _could_ add an explicit setter to this CP, and that way I think that (really tiny) edge case would be avoided...

tomnez:
@vladucu @andrewpye updated - also I left the `controllers/test/index.js` file in there because I have found myself repeatedly re-creating it to test new components. Any opposition to this?

addon/components/polaris-popover.js
tomnez: `onOpen` doesn't overwrite anything. `verticalPosition` is still a CP, so if `preferredPosition` dynamically changes, it will re-set `verticalPosition` to the new value, and then when `onOpen` runs it will _not_ compute a new value for `verticalPosition` because it only computes a new value if `preferredPosition` is set to `"mostSpace"`.

andrewpye: After `set`ting `verticalPosition` on [line 161](https://github.com/smile-io/ember-polaris/blob/3915ec85eb771e0441db38b6de7f8c5d8762c9c3/addon/components/polaris-popover.js#L161), my understanding is that `verticalPosition` will _no longer be a CP_ - the `set` will replace the entire CP with the literal value that you're setting it to, unless the CP has its own `set` defined as well. Think of it like this, where `someObject` is the result of the `computed` call:

```js
  ... other props...

  myComputed: someObject,

  ... other stuff...
```

If you then `this.set('myComputed', 'something else')`, you'll _replace `someObject` with the new string literal_. The CP "contained" in `someObject` will be torn down, and will no longer be computed on dependency changes. From then on, `this.get('myComputed')` will _always_ return `'something else'`.

You can work around this by defining the CP as

```js
myComputed: computed('dep1', 'dep2', {
  get() {
    return 'whatever';
  },

  set(key, value) {
    return value;
  }
),
```

The `set` part ensures that the CP remains after a `this.set('myComputed', 'something else')`, and the value will still update after `dep1` or `dep2` change...

addon/templates/components/polaris-popover.hbs
```diff
@@ -1,4 +1,6 @@
 {{#basic-dropdown
+  verticalPosition=verticalPosition
+  onOpen=(action 'onOpen')
```
andrewpye: Double-quotes in template files pls 😉

tests/integration/components/polaris-date-picker-test.js
```diff
@@ -339,7 +339,7 @@ test('it does not fire actions when disabled days are clicked', function(assert)
     year: YEAR,
     selected: null,
     disableDatesAfter: DISABLE_AFTER,
-    onChange: (selected) => {
+    onChange: () => {
```
andrewpye: Not sure you need the colon here actually

tests/integration/components/polaris-popover-test.js
```diff
@@ -101,3 +103,40 @@ test('it renders the correct HTML with sectioned attribute', function(assert) {
...
+  // close the popover before resetting position
+  click(activatorSelector);
+
+  this.set('preferredPosition', 'below');
+
+  click(activatorSelector);
+
+  const contentBelow = document.querySelector(popoverContentBelowSelector);
+  assert.ok(contentBelow, 'renders popover content below the trigger');
+});
```
andrewpye: Love that you got this working 😍

PR #128, opened 2018-05-15
Update polaris-card/section to v2.0.0

andrewpye:
Adds the [new `fullWidth` property](https://github.com/Shopify/polaris/compare/v1.12.4...v2?diff=split#diff-22b2ce0337c39af9c73a2bf0d7ee1569R12) to `polaris-card/section` to allow removing padding from card sections.


tomnez:
@andrewpye should we add this as an attribute in the docs examples?

andrewpye:
@tomnez good shout on the docs update, thanks! 👍

PR #132, opened 2018-05-24
Move to npm org tweaks

vladucu:
### What's here

- updates to use scoped package in both `package.json` and `index.js`
- updates exports/imports to use scoped package

NOTE: Until now everything worked because we didn't changed `index.js` name, with this we will break imports and stuff....however we need to do this, since otherwise newly generated stuff will use scoped name, but import won't resolve because it will still use un-scoped name

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - "4"
+  - "6"
 
 sudo: false
 dist: trusty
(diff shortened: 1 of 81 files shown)


andrewpye:
eep, failed on Travis... _I take back what I just said_ 😜

vladucu:
meh, I guess I was the first one that run into this
waiting for travis...doesn't seems like our issue

andrewpye:
Looks like a node version incompatibility with `broccoli-svg-optimizer` 😬 

image

vladucu:
yeap, but we didn't touched that and it worked so far....checking more

andrewpye:
Yeah... thanks 🙂

PR #135, opened 2018-05-29
Fix component documentation links

andrewpye:
Ensures component documentation links are always relative


andrewpye:
Travis is failing because of some issues... 😞

vladucu:
yeah, there have been temporary registry issues since Friday

PR #136, opened 2018-05-29
Fix a few small issues in `polaris-tag`

andrewpye:
Fixes a few small things I found were a bit iffy with `polaris-tag`:

- uses `cancel-small` for the remove button icon instead of `cancel`
- moves `handleMouseUpByBlurring` from the tag component itself to the `remove` button
- ensures the `remove` button's `aria-label` always contains the latest tag text


andrewpye:
Travis errors are due to yarn issues, going to get this out the door because tests pass locally 😉

PR #139, opened 2018-06-12
Make `polaris-spinner` respect the `color` property

### LINKED ISSUE #116: Spinner does not obey `color` attribute
The spinner SVG has a `fill` attribute on it which causes the image to always be grey. We need to remove this attribute as we do in `polaris-icon`, so we should probably abstract this behaviour into a mixin and reuse it in `polaris-spinner`. However, the latter is tagless, so we'll need to find some reliable way of grabbing the spinner's SVG element.

andrewpye:
Fixes #116 by abstracting out the SVG `fill` attribute processing from `polaris-icon` into a mixin and applying it to the SVG rendered by `polaris-spinner` as well.

### DIFF
--- a/addon/mixins/components/svg-handling.js
+++ b/addon/mixins/components/svg-handling.js
@@ -35,6 +35,12 @@ export default Mixin.create({
 
   /**
    * Volatile CP to find the SVG element for this component.
+   * This is marked as `volatile` so that `this.get('svgElement')`
+   * essentially becomes an alias for the `document.querySelector`
+   * call, meaning that each time we try to access it, we'll get
+   * an up-to-date indication of whether the element is still
+   * in the DOM. Without the `volatile` modifier, we could for
+   * example end up using a cached element that's no longer in the DOM.
    *
    * @property svgElement
    * @type {DOMNode}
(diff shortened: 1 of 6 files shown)


addon/mixins/components/svg-handling.js
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Volatile CP to find the SVG element for this component.
+   *
+   * @property svgElement
+   * @type {DOMNode}
+   */
+  svgElement: computed(function() {
+    return document.querySelector(`#${ this.get('svgElementId') }`);
+  }).volatile(),
```
vladucu: does this needs to be volatile?

andrewpye: I figured that this way, it's basically the same as making `document.querySelector` call inline, just a bit cleaner. This means that we're checking against the DOM each time we try using `svgElement`, rather than caching an element that could get removed from the DOM at some point (e.g. if the icon is wrapped in an `if`-condition etc.). It shouldn't affect anything as things currently are, but this way just seemed a bit safer for the future.

If you don't like the `volatile` modifier, I can just make the `querySelector` call inline instead... this just seemed like a nice way of wrapping up finding the SVG element 😛

vladucu: k, yeah, that makes sense...I didn't thought we need it right now for our use-cases...that's why I asked
I'm fine with leaving it for now...maybe add a comment for future-us 😄

vladucu:
[Investigate spinner coloring bug](https://trello.com/c/EzNKOc0g/143-investigate-spinner-coloring-bug)

PR #141, opened 2018-06-22
Add data table component

andrewpye:
Depends on #142 in order to be able to render arbitrary components in table cells.

Adds full implementation of the Polaris [data table](https://polaris.shopify.com/components/lists-and-tables/data-table) component:

Reproducing the demos on the Shopify component documentation page:

Data table with footer:
polaris-data-table-footer

Data table with sortable columns (when the sorting updates without clicking it's because I pressed `Enter` while the column heading was focussed):
polaris-data-table-sortable

### DIFF
--- a/addon/components/polaris-data-table.js
+++ b/addon/components/polaris-data-table.js
@@ -375,6 +375,10 @@ export default Component.extend({
   },
 
   debouncedHandleResize() {
+    if (this.get('isDestroying') || this.get('isDestroyed')) {
+      return;
+    }
+
     let { footerContent, truncate } = this.getProperties('footerContent', 'truncate');
     let collapsed = this.get('table.scrollWidth') > this.get('dataTable.offsetWidth');
 
@@ -395,6 +399,10 @@ export default Component.extend({
   },
 
   scrollListener() {
+    if (this.get('isDestroying') || this.get('isDestroyed')) {
+      return;
+    }
+
     this.setProperties(this.calculateColumnVisibilityData(this.get('collapsed')));
   },
 
@@ -479,6 +487,10 @@ export default Component.extend({
 
       // TODO: use run loop instead of `requestAnimationFrame` here?
       requestAnimationFrame(() => {
+        if (this.get('isDestroying') || this.get('isDestroyed')) {
+          return;
+        }
+
         this.setProperties(this.calculateColumnVisibilityData(this.get('collapsed')));
       });
     },
--- a/addon/components/polaris-data-table.js
+++ b/addon/components/polaris-data-table.js
@@ -483,8 +483,39 @@ export default Component.extend({
       });
     },
 
-    defaultOnSort() {
-      // TODO: implement this.
+    defaultOnSort(headingIndex) {
+      let {
+        onSort,
+        truncate,
+        defaultSortDirection = 'ascending',
+        initialSortColumnIndex,
+        sortDirection,
+        sortedColumnIndex = initialSortColumnIndex,
+      } = this.getProperties('onSort', 'truncate', 'defaultSortDirection', 'initialSortColumnIndex', 'sortDirection', 'sortedColumnIndex');
+      let newSortDirection = defaultSortDirection;
+      if (sortedColumnIndex === headingIndex) {
+        newSortDirection = sortDirection === 'ascending' ? 'descending' : 'ascending';
+      }
+
+      this.setProperties({
+        sorted: true,
+        sortDirection: newSortDirection,
+        sortedColumnIndex: headingIndex,
+      });
+
+      scheduleOnce('afterRender', () => {
+        if (onSort) {
+          onSort(headingIndex, newSortDirection);
+          if (!truncate) {
+            let preservedScrollPosition = {
+              left: this.scrollContainer.scrollLeft,
+              top: window.scrollY,
+            };
+            this.set('preservedScrollPosition', preservedScrollPosition);
+            this.handleResize();
+          }
+        }
+      });
     },
   }
 });
--- a/addon/components/polaris-data-table.js
+++ b/addon/components/polaris-data-table.js
@@ -2,9 +2,10 @@ import Component from '@ember/component';
 import { computed } from '@ember/object';
 import { isBlank, isPresent, isNone } from '@ember/utils';
 import { htmlSafe } from '@ember/string';
-import { scheduleOnce, debounce } from '@ember/runloop';
+import { scheduleOnce } from '@ember/runloop';
 import { assign } from '@ember/polyfills';
 import ContextBoundEventListenersMixin from 'ember-lifeline/mixins/dom';
+import ContextBoundTasksMixin from 'ember-lifeline/mixins/run';
 import layout from '../templates/components/polaris-data-table';
 
 function elementLookup(selector) {
@@ -78,7 +79,7 @@ function getPrevAndCurrentColumns(tableData, columnData) {
  * Polaris data table component.
  * See https://polaris.shopify.com/components/lists-and-tables/data-table
  */
-export default Component.extend(ContextBoundEventListenersMixin, {
+export default Component.extend(ContextBoundEventListenersMixin, ContextBoundTasksMixin, {
   classNameBindings: ['collapsed:Polaris-DataTable--collapsed'],
 
   layout,
@@ -359,14 +360,10 @@ export default Component.extend(ContextBoundEventListenersMixin, {
 
   handleResize() {
     // This is needed to replicate the React implementation's `@debounce` decorator.
-    debounce(this, 'debouncedHandleResize', 0);
+    this.debounceTask('debouncedHandleResize', 0);
   },
 
   debouncedHandleResize() {
-    if (this.get('isDestroying') || this.get('isDestroyed')) {
-      return;
-    }
-
     let { footerContent, truncate } = this.getProperties('footerContent', 'truncate');
     let collapsed = this.get('table.scrollWidth') > this.get('dataTable.offsetWidth');
 
--- a/addon/components/polaris-data-table.js
+++ b/addon/components/polaris-data-table.js
@@ -1,6 +1,6 @@
 import Component from '@ember/component';
 import { computed } from '@ember/object';
-import { isBlank, isPresent } from '@ember/utils';
+import { isBlank, isPresent, isNone } from '@ember/utils';
 import { htmlSafe } from '@ember/string';
 import { scheduleOnce, debounce } from '@ember/runloop';
 import { assign } from '@ember/polyfills';
@@ -490,8 +490,9 @@ export default Component.extend({
         defaultSortDirection = 'ascending',
         initialSortColumnIndex,
         sortDirection,
-        sortedColumnIndex = initialSortColumnIndex,
+        sortedColumnIndex,
       } = this.getProperties('onSort', 'truncate', 'defaultSortDirection', 'initialSortColumnIndex', 'sortDirection', 'sortedColumnIndex');
+      sortedColumnIndex = isNone(sortedColumnIndex) ? initialSortColumnIndex : sortedColumnIndex;
       let newSortDirection = defaultSortDirection;
       if (sortedColumnIndex === headingIndex) {
         newSortDirection = sortDirection === 'ascending' ? 'descending' : 'ascending';
@@ -508,7 +509,7 @@ export default Component.extend({
           onSort(headingIndex, newSortDirection);
           if (!truncate) {
             let preservedScrollPosition = {
-              left: this.scrollContainer.scrollLeft,
+              left: this.get('scrollContainer').scrollLeft,
               top: window.scrollY,
             };
             this.set('preservedScrollPosition', preservedScrollPosition);
--- a/addon/components/polaris-data-table.js
+++ b/addon/components/polaris-data-table.js
@@ -4,12 +4,8 @@ import { isBlank, isPresent, isNone } from '@ember/utils';
 import { htmlSafe } from '@ember/string';
 import { scheduleOnce, debounce } from '@ember/runloop';
 import { assign } from '@ember/polyfills';
+import ContextBoundEventListenersMixin from 'ember-lifeline/mixins/dom';
 import layout from '../templates/components/polaris-data-table';
-
-const eventHandlerParams = [
-  [ window, 'resize', 'handleResize' ],
-  [ window, 'scroll', 'scrollListener', true ],
-];
 
 function volatileElementLookup(selector) {
   return computed(function() {
@@ -82,7 +78,7 @@ function getPrevAndCurrentColumns(tableData, columnData) {
  * Polaris data table component.
  * See https://polaris.shopify.com/components/lists-and-tables/data-table
  */
-export default Component.extend({
+export default Component.extend(ContextBoundEventListenersMixin, {
   classNameBindings: ['collapsed:Polaris-DataTable--collapsed'],
 
   layout,
@@ -266,13 +262,6 @@ export default Component.extend({
   totalsRowHeading: 'Totals',
 
   /**
-   * @property eventHandlerParams
-   * @type {Array[]}
-   * @private
-   */
-  eventHandlerParams,
-
-  /**
    * @property dataTable
    * @type {HTMLElement}
    * @private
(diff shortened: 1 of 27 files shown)


andrewpye:
[Implement Polaris data-table component](https://trello.com/c/jer0zMml/141-implement-polaris-data-table-component)

addon/templates/components/polaris-data-table/navigation.hbs
```diff
@@ -0,0 +1,19 @@
+{{polaris-button
+  plain=true
+  icon="chevron-left"
+  disabled=currentColumn.isScrolledFarthestLeft
+  accessibilityLabel="Scroll table left one column"
+  onClick=(action navigateTableLeft)
+}}
+
+{{#each columnVisibilityData as |column|}}
+  <div class="Polaris-DataTable__Pip {{if column.isVisible "Polaris-DataTable__Pip--visible"}}"></div>
```
tomnez: Maybe single-quote the inner string for slightly easier readability? No biggie if we prefer double quotes even for nested strings in templates, but just a thought.

tests/integration/components/polaris-data-table-test.js
```diff
@@ -0,0 +1,149 @@
...
+        sortable=sortable
+        initialSortColumnIndex=4
+      }}
+    `);
+    const fifthHeadingCell = findAll('th.Polaris-DataTable__Cell')[5];
+
+    assert.ok(fifthHeadingCell.classList.contains('Polaris-DataTable__Cell--sorted'));
+  });
+
+  test('it accepts both text and component definitions as cell contents', async function(assert) {
```
tomnez: 💯 💯 💯

addon/components/polaris-data-table/cell.js
```diff
@@ -0,0 +1,192 @@
...
+   * @public
+   */
+  truncate: false,
+
+  /**
+   * @property presentational
+   * @type {boolean}
+   * @public
+   */
+  presentational: false,
```
tomnez: There are some inconsistencies in these files on whether or not there's a `@default` added - I think we want `@default` notation for any non-null attribute, yes?

addon/components/polaris-data-table.js
```diff
@@ -0,0 +1,534 @@
...
+
+    return {
+      columnVisibilityData: [],
+      previousColumn: undefined,
+      currentColumn: undefined,
+    };
+  },
+
+  handleResize() {
+    // This is needed to replicate the React implementation's `@debounce` decorator.
```
tomnez: 👍

addon/components/polaris-data-table.js
```diff
@@ -0,0 +1,534 @@
...
+      let { scrollContainer, currentColumn, previousColumn } = this.getProperties('scrollContainer', 'currentColumn', 'previousColumn');
+
+      if (direction === 'right' && currentColumn) {
+        scrollContainer.scrollLeft = currentColumn.rightEdge;
+      } else if (previousColumn) {
+        scrollContainer.scrollLeft =
+          previousColumn.leftEdge < 10 ? 0 : previousColumn.leftEdge;
+      }
+
+      // TODO: use run loop instead of `requestAnimationFrame` here?
```
tomnez: Seems like `requestAnimationFrame` is just an overkill way of quickly calculating precise position values of columns? I don't know if swapping for run loop would accomplish whatever the intention of `requestAnimationFrame` here is, maybe just leave as-is unless you're certain that we need to throw this into one of the run loop cycles.

addon/components/polaris-data-table/cell.js
```diff
@@ -0,0 +1,192 @@
...
+  /**
+   * @property contentType
+   * @type {String}
+   * @public
+   */
+  contentType: null,
+
+  /**
+   * @property fixed
+   * @type {boolean}
```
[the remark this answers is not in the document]
andrewpye: Yeah... and across the whole addon 😬 If we ever get the stage of using the JSDoc to generate documentation we should definitely standardise this, but at this point I'm not that fussed. Curious as to @vladucu's thoughts on this...

addon/templates/components/polaris-data-table/cell.hbs
```diff
@@ -0,0 +1,72 @@
...
+      {{#if sortable}}
+        <th
+          scope="col"
+          class={{cellClassNames}}
+          style={{style}}
+          role="button"
+          aria-sort={{sortDirection}}
+          aria-label={{sortAccessibilityLabel}}
+          tabindex="0"
+          {{action onSort}}
```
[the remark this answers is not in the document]
andrewpye: `onclick` will pass an event object to the action while this usage doesn't - for the keydown handler we need that event to determine whether it was the Enter key that was pressed, whilst for the click handler we don't need it. I prefer to not pass that object if possible since it's something of an implementation detail and shouldn't really be exposed to the outside world IMO.

addon/templates/components/polaris-data-table/cell.hbs
```diff
@@ -0,0 +1,72 @@
...
+          onkeydown={{action "onKeyDown"}}
+        >
+          {{#if (eq contentType "text")}}
```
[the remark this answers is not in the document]
andrewpye: Yes, this whole template could be DRY'd up a lot, but I want to keep it as similar to the React implementation as possible to make future updates easier to pick out

addon/templates/components/polaris-data-table/navigation.hbs
andrewpye: Fair point, never single-quoted in template files before. I'm OK with this as is, but if both you and @vladucu would prefer it changed I'm happy to change it, no biggie as you say 😉

addon/components/polaris-data-table/total.js
```diff
@@ -0,0 +1,56 @@
...
+import { computed } from '@ember/object';
+import layout from '../../templates/components/polaris-data-table/total';
+
+export default Component.extend({
+  tagName: '',
+
+  layout,
+
+  /**
+   * TODO: make this support passing a component to render instead of just text/numbers.
```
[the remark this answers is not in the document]
andrewpye: Good catch, I rushed through adding `render-content` yesterday afternoon and missed this 👍

addon/components/polaris-data-table.js
andrewpye: My understanding is that `requestAnimationFrame` helps prevent jankiness on scrolling, resizing etc. by limiting the amount of work done to only that which can be fitted between two consecutive frame renders. With this in mind, using it here makes some kind of sense, although I'm slightly confused as to why it's not being used anywhere else (I'd expect the resize handler to be a candidate for its use, for example).

That being said, I'm not sure if using it directly here creates the possibility of interfering with the Ember run loop. The interesting-looking [`ember-run-raf`](https://github.com/runspired/ember-run-raf) addon seems to be advocating "using `run.scheduleOnce` + `requestAnimationFrame`" in combination, so maybe we should be doing that combo here, not sure...

addon/components/polaris-data-table/cell.js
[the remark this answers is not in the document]
tomnez: He'd probably say something like [this](https://github.com/smile-io/ember-polaris/pull/104#discussion_r175738241).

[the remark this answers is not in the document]
andrewpye: Yeah maybe 😛 We currently have 22 `{Boolean}`s and 54 `{boolean}`s, so clearly we'll need to tidy those up at some point 😬 I think it's a big enough job to be done separately from this PR 🤷‍♂️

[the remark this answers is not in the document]
tomnez: Yeah I just meant change the files in this PR, not across the addon. No reason to continue a bad habit just because it's already been done elsewhere.

vladucu:
sorry for the delay on this, getting to it in a bit...

addon/components/polaris-data-table.js
tomnez: Oh wow I didn't realize there was an addon that tied the two together. If it's owned/authored by runspired then I trust it, but again for this purpose I don't think it's really an urgent need. It's up to you if you want to experiment with it but I think I'm fine to leave it as-is.

andrewpye: Yeah agreed, that guy knows what he's doing (though I bet he wishes he was @vladucu 😉)! I wasn't intending to pull in the addon here though, just found it the other day and thought it was interesting.

addon/templates/components/polaris-data-table/navigation.hbs
vladucu: I've usually used single-quotes in cases like this just to avoid confusion and be easier to read

addon/templates/components/polaris-data-table.hbs
```diff
@@ -0,0 +1,72 @@
...
+          }}
+        {{/each}}
+      </tbody>
+
+      {{#if footerContent}}
+        <tfoot class="Polaris-DataTable__TableFoot">
+          <tr>
+            {{polaris-data-table/cell
+              footer=true
+              height=heights.lastObject
```
vladucu: can we pass `heights` here and compute this inside as we did with other react-renderers -> components?

addon/components/polaris-data-table.js
```diff
@@ -0,0 +1,535 @@
...
+  );
+  return {previousColumn, currentColumn};
+}
+
+/**
+ * Polaris data table component.
+ * See https://polaris.shopify.com/components/lists-and-tables/data-table
+ */
+export default Component.extend({
+  classNameBindings: ['collapsed:Polaris-DataTable--collapsed'],
```
vladucu: Polaris adds a `TableWrapper` class here, I reckon there's no styles and not used...maybe we should still add it?

addon/components/polaris-data-table.js
```diff
@@ -0,0 +1,535 @@
...
+    if (footerContent) {
+      let footerCellHeight = rows[rows.length - 1].children[0].clientHeight;
+      heights = [footerCellHeight];
+    }
+
+    return heights;
+  },
+
+  addEventHandlers() {
+    this.get('eventHandlerParams').forEach(([ target, eventName, callbackName, ...rest ]) => {
```
vladucu: we already have `ember-lifeline` as a dependency...we should use it for adding event listeners....there's a mixin which will also do the cleanup for us and I don't think you have to guard against the component being destroyed

addon/components/polaris-data-table.js
```diff
@@ -0,0 +1,535 @@
...
+
+  setHeightsAndScrollPosition() {
+    this.set('heights', this.tallestCellHeights());
+
+    scheduleOnce('afterRender', this, this.resetScrollPosition);
+  },
+
+  calculateColumnVisibilityData(collapsed) {
+    if (collapsed) {
+      let headerCells = this.get('table').querySelectorAll(
```
vladucu: we can make this a one-liner?

addon/components/polaris-data-table.js
```diff
@@ -0,0 +1,535 @@
...
+      leftEdge < tableLeftVisibleEdge && rightEdge > tableRightVisibleEdge;
+    let isVisible =
+      isCompletelyVisible || leftEdgeIsVisible || rightEdgeIsVisible;
+    if (isVisible) {
+      tableData.firstVisibleColumnIndex = Math.min(
+        firstVisibleColumnIndex,
+        index,
+      );
+    }
+    return {leftEdge, rightEdge, isVisible};
```
vladucu: ni: spaces

addon/components/polaris-data-table.js
```diff
@@ -0,0 +1,535 @@
...
+    this.setProperties(this.calculateColumnVisibilityData(this.get('collapsed')));
+  },
+
+  tallestCellHeights() {
+    let { footerContent, truncate, heights } = this.getProperties('footerContent', 'truncate', 'heights');
+    let rows = Array.from(this.get('table').getElementsByTagName('tr'));
+
+    if (!truncate) {
+      return (heights = rows.map((row) => {
+        let fixedCell = row.children[0];
```
vladucu: react uses `childNodes`, any reason we don't follow?

addon/components/polaris-data-table.js
vladucu: add a note here maybe to check `ember-run-raf` when testing this in the real world?

addon/components/polaris-data-table.js
```diff
@@ -0,0 +1,535 @@
...
+   * @private
+   */
+  scrollContainer: volatileElementLookup('.Polaris-DataTable__ScrollContainer').readOnly(),
+
+  /**
+   * @property table
+   * @type {HTMLElement}
+   * @private
+   */
+  table: volatileElementLookup('.Polaris-DataTable__Table').readOnly(),
```
vladucu: I don't think these element lookups need to be volatile

tests/integration/components/polaris-data-table-test.js
```diff
@@ -0,0 +1,149 @@
+import { helper } from '@ember/component/helper';
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+import { findAll, find } from 'ember-native-dom-helpers';
```
vladucu: we should use the new find, findAll from `@ember/test-helpers`....we can probably start deprecating native-dom helpers

vladucu: not saying this should not go out because of that^

we should also use qunit-dom for assertion, from what I've seen much nicer error messages

addon/templates/components/polaris-data-table.hbs
andrewpye: I don't see any benefit to doing this, and I'd also prefer not to since the `Cell` component is a port of the React implementation which takes `height` instead of `heights`

addon/components/polaris-data-table.js
andrewpye: Do you see that class being applied in actual usage? I checked this and nothing seems to end up getting applied to the element:

image

I guess there's no defined style for it in the Polaris source so it gets stripped out or something 🤷‍♂️

addon/components/polaris-data-table.js
andrewpye: This is just a tweak from what's in the [React source](https://github.com/Shopify/polaris/blob/master/src/components/DataTable/DataTable.tsx#L337), I wanted to keep the two implementations similar to make changes more visible in the future...

addon/templates/components/polaris-data-table.hbs
vladucu: ah, you're right...thought it's some other component

addon/components/polaris-data-table.js
vladucu: yes most likely it's stripped out

addon/components/polaris-data-table.js
vladucu: well it's still the same thing just one-line no?

andrewpye: Yeah, I literally just copied the React code and changed the bits that needed it. Can go through and tweak the styling of all the copied code if you want (will fix the "spaces" nit comment on line 51 as well)

vladucu: however you feel too...we wont' be able to compare code through a git-diff style anyway....so I'd be for these small tweaks, but flexible

addon/components/polaris-data-table.js
andrewpye: Yep, `childNodes` includes a bunch of extra `text` nodes in Ember land which aren't present in the React world (spent a lil while tracking down why this wasn't behaving properly and that turned out to be the cause) 🤷‍♂️

addon/components/polaris-data-table.js
andrewpye: I guess you're right actually 👍

addon/components/polaris-data-table.js
andrewpye: From what I can see, `ember-run-raf` changes the _entire_ run loop to use `requestAnimationFrame` rather than adding the option to e.g. schedule on an animation frame queue, so probs not something we want to pull in to the host app 😬

addon/components/polaris-data-table.js
andrewpye: @vladucu refactored to use `ember-lifeline` for event handlers and debouncing, but it [cant be used for `afterRender` scheduling](https://github.com/ember-lifeline/ember-lifeline#a-word-about-the-afterrender-queue) so I've left that as is.

andrewpye:
@vladucu changes pushed, ready for 👀 pt. deux 😅

andrewpye:
(tests are currently failing on Travis due to package issues 😞They all pass locally!)

addon/components/polaris-data-table/cell.js
```diff
@@ -188,6 +209,17 @@ export default Component.extend({
     return `caret-${ sortDirection === 'ascending' ? 'up' : 'down' }`;
   }).readOnly(),
 
+  updateAccessibilityLabel() {
+    let cellElement = document.querySelector(`#${ this.get('cellElementId') }`);
+    this.set('sortAccessibilityLabel', `Sort by ${cellElement.textContent.trim().toLowerCase()}`);
```
vladucu: nit: spaces

PR #142, opened 2018-06-25
Support React-style content rendering

andrewpye:
Quick little enhancement that allows us to render both "simple" values (e.g. `componentContent="some text"`) and component definitions (passed as e.g. `componentContent=(component "my-cool-component" color="pink")`) using the same one-liner, `{{render-content componentContent}}`.

This opens the door for us to be able to easily support passing child content to `ember-polaris` components without having to go down the route we've taken before of adding e.g. separate `label` and `labelComponent` properties to Polaris components.

### DIFF
--- a/addon/components/render-content.js
+++ b/addon/components/render-content.js
@@ -1,0 +1,18 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import layout from '../templates/components/render-content';
+
+export default Component.extend({
+  tagName: '',
+
+  layout,
+
+  content: null,
+
+  contentIsComponent: computed('content', function() {
+    let contentConstructorName = this.get('content.constructor.name') || '';
+    return contentConstructorName.indexOf('ComponentDefinition') > -1;
+  }).readOnly(),
+}).reopenClass({
+  positionalParams: ['content'],
+});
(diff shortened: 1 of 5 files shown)


addon/components/render-content.js
```diff
@@ -0,0 +1,18 @@
...
+  layout,
+
+  content: null,
+
+  contentIsComponent: computed('content', function() {
+    let contentConstructorName = this.get('content.constructor.name') || '';
+    return contentConstructorName.indexOf('ComponentDefinition') > -1;
+  }).readOnly(),
+}).reopenClass({
+  positionalParams: ['content'],
```
tomnez: Ooh 🥇

tests/integration/components/render-content-test.js
```diff
@@ -0,0 +1,36 @@
...
+    this.owner.register('component:my-component', Component.extend({
+      classNames: ['my-test-component'],
+      layout: hbs`{{text}}`,
+    }));
+  });
+
+  test('it renders simple content correctly', async function(assert) {
+    await render(hbs`
+      <div id="render-content-test">
+        {{render-content "blah"}}
```
tomnez: I've never seen a test with such an attitude 😆

tests/integration/components/render-content-test.js
```diff
@@ -0,0 +1,36 @@
...
+      </div>
+    `);
+
+    assert.equal(find('#render-content-test').textContent.trim(), 'blah');
```
tomnez: Only nitpicky suggestion would be to use `const`s for the text values that you're setting and checking, but this is a small and simple test so not a big deal.

PR #145, opened 2018-06-27
Polaris text field

tomnez:
### Overview

Implements `polaris-text-field`.

Text field [Polaris docs](https://polaris.shopify.com/components/forms/text-field)
Text field [Github component](https://github.com/Shopify/polaris/tree/master/src/components/TextField)

Notes:

- Label Action: The `labelAction` property was implemented with the `content`, `accessibilityLabel`, and `onAction` attributes. `external`, `id`, and `url` were omitted - My thinking was that the parent component could link to external URLs if needed (via `onAction`) rather than forcing users to use one specific way to link to external URLs since there are possible vulnerabilities that we don't need to make ourselves responsible for.

- Chatted with @andrewpye about whether to use Ember's `{{textarea}}` and `{{input}}` helpers or use actual HTML elements. Ember doesn't have built-in bindings for a lot of the attributes Polaris implements, so we went with the latter. There are some methods in `polaris-text-field.js` to bind the `value` to the value of the textarea/inputs. If this isn't ideal I can swap it for using the Ember helpers, we'd just lose out on some of the `on{event}` attributes.

### Demos
(FYI this is the dummy app so icons aren't rendering)

**with prefix and suffix**
prefix-suffix

**placeholder**
placeholder

**numeric type**
number

**multiline**
multiline

**with help text**
helptext

**error state**
error

**disabled state**
disabled

**with action label**
action label

**with `connectedLeft` and `connectedRight` attributes**
connected-left-right

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ Below is a categorised list of the components available in `ember-polaris`. Clic
 - [Form layout](./docs/form-layout.md#form-layout)
 - [Radio button](./docs/radio-button.md#radio-button)
 - [Tag](./docs/tag.md#tag)
+- [Text field](./docs/text-field.md#text-field)
 
 #### Lists
 - [Data table](./docs/data-table.md#data-table)
--- a/docs/text-field.md
+++ b/docs/text-field.md
@@ -1,0 +1,79 @@
+[< Back to Components List](../README.md#components)
+
+## Text field
+
+`polaris-text-field` implements the [Polaris Text field component](https://polaris.shopify.com/components/forms/text-field).
+
+### Examples
+
+Basic usage:
+
+```hbs
+{{polaris-text-field
+  label="First name"
+  placeholder="e.g. Bob"
+  value=value
+  onChange=(action "handleChange")
+}}
+```
+
+Numeric field with hidden label:
+
+```hbs
+{{polaris-text-field
+  label="Quantity"
+  labelHidden=true
+  type="number"
+  value=value
+  onChange=(action "handleChange")
+}}
+```
+
+Text field with label action:
+
+```hbs
+{{polaris-text-field
+  value=value
+  onChange=(action "handleChange")
+  labelAction=(hash
+    text="Forgot password?"
+    onAction=(action "resetPassword")
+  )
+}}
+```
+
+Text field with prefix, suffix, and help text:
+
+```hbs
+{{polaris-text-field
+  label="Price"
+  prefix="$"
+  suffix="USD"
+  helpText="Some help text"
+  value=value
+}}
+```
+
+Text field with connected fields:
+
+```hbs
+{{polaris-text-field
+  value=value
+  connectedLeft="Connection as string"
+  connectedRight=(component
+    "polaris-button"
+    text="Connection as component"
+    onClick=(action "myAction")
+  )
+}}
+```
+
+Text field with multiline:
+
+```hbs
+{{polaris-text-field
+  multiline=true
+  value=value
+  onChange=(action "handleChange")
+}}
+```


addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+    let { value, disabled, readOnly, error, multiline, focus } = this.getProperties('value', 'disabled', 'readOnly', 'error', 'multiline', 'focus');
+
+    let valueClass = value ? 'Polaris-TextField--hasValue' : '';
+    let disabledClass = disabled ? 'Polaris-TextField--disabled' : '';
+    let readOnlyClass = readOnly ? 'Polaris-TextField--readOnly' : '';
+    let errorClass = error ? 'Polaris-TextField--error' : '';
+    let multilineClass = multiline ? 'Polaris-TextField--multiline' : '';
+    let focusClass = focus ? 'Polaris-TextField--focus' : '';
+
+    return `Polaris-TextField ${ valueClass } ${ disabledClass } ${ readOnlyClass } ${ errorClass } ${ multilineClass } ${ focusClass }`;
```
andrewpye: lol, this gives you one long chunk of trailing whitespace if none of those flags are truthy 😛 How do you feel about pushing the classes onto an array and returning `classes.join(' ')`?

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+   * Callback when focus is removed
+   *
+   * @property onBlur
+   * @public
+   * @type {Function}
+   * @default noop
+   */
+  onBlur() {},
+
+  addValueListener() {
```
andrewpye: This function should come after CPs

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+
+    let valueClass = value ? 'Polaris-TextField--hasValue' : '';
+    let disabledClass = disabled ? 'Polaris-TextField--disabled' : '';
+    let readOnlyClass = readOnly ? 'Polaris-TextField--readOnly' : '';
+    let errorClass = error ? 'Polaris-TextField--error' : '';
+    let multilineClass = multiline ? 'Polaris-TextField--multiline' : '';
+    let focusClass = focus ? 'Polaris-TextField--focus' : '';
+
+    return `Polaris-TextField ${ valueClass } ${ disabledClass } ${ readOnlyClass } ${ errorClass } ${ multilineClass } ${ focusClass }`;
+  }),
```
andrewpye: Make all of these internal-only CPs `.readOnly()`

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+    }
+
+    return describedBy.join(' ');
+  }),
+
+  ariaLabelledBy: computed('id', function() {
+    return `${ this.get('id') }Label`;
+  }),
+
+  ariaInvalid: computed('error', function() {
```
andrewpye: This can just be `bool('error').readOnly()`

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+    return !!this.get('error');
+  }),
+
+  inputType: computed('type', function() {
+    let type = this.get('type');
+
+    return type === 'currency' ? 'text' : type;
+  }),
+
+  input: computed('multiline', function() {
```
andrewpye: Are you using this anywhere?

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+    let multiline = this.get('multiline');
+
+    return multiline ? 'textarea' : 'input';
+  }),
+
+  minimumLines: computed('multiline', function() {
+    let multiline = this.get('multiline');
+
+    return (typeof multiline === 'number') ? multiline : 1;
```
andrewpye: Use `typeOf` from `@ember/utils`

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+  }),
+
+  heightStyle: computed('height', 'multiline', function() {
+    let { height, multiline } = this.getProperties('height', 'multiline');
+
+    return (height && multiline) ? htmlSafe(`height: ${ height }px`) : '';
```
andrewpye: Is this better being `null` than an empty string?

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+    });
+  },
+
+  didReceiveAttrs() {
+    this._super(...arguments);
+
+    let { input, focused } = this.getProperties('input', 'focused');
+
+    if (input && focused) {
+      input.focus();
```
andrewpye: Isn't `input` a string here (either `'input'` or `'textarea'`)? 😕

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+    }
+  },
+
+  willDestroyElement() {
```
andrewpye: This and `destroy` can go using `ember-lifeline` as described in other comments 🙌

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+        id,
+        onChange,
+        value,
+        step,
+        min,
+        max,
+      } = this.getProperties('id', 'onChange', 'value', 'step', 'min', 'max');
+
+      step = step || 1;
+      min = min || -Infinity;
```
andrewpye: This is buggy since if `min` is specified as `0` you'll end up with `-Infinity` here. Maybe use `isPresent` or `isNone` here instead.

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+      max = max || Infinity;
```
andrewpye: Ditto here re. `0` bug

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+      let numericValue = value ? parseFloat(value) : 0;
+
+      if (isNaN(numericValue)) {
+        return;
+      }
+
+      // Making sure the new value has the same length of decimal places as the
+      // step / value has.
+      let decimalPlaces = Math.max(dpl(numericValue), dpl(step));
+      let newValue = Math.min(max, Math.max(numericValue + steps * step, min));
```
andrewpye: These Polaris guys are pretty good at what they do 💯

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+
+      onChange(String(newValue.toFixed(decimalPlaces)), id);
```
andrewpye: I know this is duplicated from the React code, but doesn't `toFixed` return a string anyway? 😕

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+    },
+
+    handleExpandingResize(height) {
+      this.set('height', height);
+    },
+
+    handleChange(e) {
+      this.get('onChange')(e, this.get('id'));
```
andrewpye: Check for `onChange` being present before calling it

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+    handleFocus() {
+      this.set('focus', true);
+    },
+
+    handleBlur() {
+      this.set('focus', false);
+    },
+
+    handleClick() {
+      let [ field ] = this.element.querySelectorAll('input, textarea');
```
andrewpye: Same here re. fetching `field` as in other comments above

addon/components/polaris-text-field/resizer.js
```diff
@@ -0,0 +1,132 @@
+import Component from '@ember/component';
+import layout from '../../templates/components/polaris-text-field/resizer';
```
andrewpye: Move this below the other `@ember` imports so the Ember imports are grouped together

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
+import Component from '@ember/component';
+import layout from '../templates/components/polaris-text-field';
+import { createUniqueIDFactory } from '../utils/other';
+import { assert } from '@ember/debug';
```
andrewpye: Group the `@ember` imports together pls

addon/components/polaris-text-field/resizer.js
```diff
@@ -0,0 +1,132 @@
...
+const REPLACE_REGEX = /[\n&<>]/g;
+
+const ENTITIES_TO_REPLACE = {
+  '&': '&amp;',
+  '<': '&lt;',
+  '>': '&gt;',
+  '\n': '<br>',
+};
+
+const replaceEntity = (entity) => {
```
andrewpye: No need for this to be defined as an arrow function 🤷‍♂️

addon/components/polaris-text-field/resizer.js
```diff
@@ -0,0 +1,132 @@
...
+  return ENTITIES_TO_REPLACE[entity] || entity;
+};
+
+export default Component.extend({
+  classNames: ['Polaris-TextField__Resizer'],
+
+  'aria-hidden': true,
```
andrewpye: Not keen on this dasherized prop name, especially as it's not bound to anything on the DOM. Make this `ariaHidden` pls 👍

EDIT: if this is _meant_ to be bound to the resizer element, you need to add an attribute binding (which can accept the `ariaHidden` name anyway)

addon/components/polaris-text-field/resizer.js
```diff
@@ -0,0 +1,132 @@
+import Component from '@ember/component';
```
andrewpye: This component's pretty genius really 🙌

addon/components/polaris-text-field/resizer.js
```diff
@@ -0,0 +1,132 @@
...
+   * @public
+   * @type {Function}
+   * @default noop
+   */
+  onHeightChange(/* height */) {},
+
+  finalContents: computed('contents', function() {
+    let contents = this.get('contents');
+
+    contents = contents
```
andrewpye: The linter doesn't complain about having the `?` and `:` on the next line? 😮

addon/components/polaris-text-field/resizer.js
```diff
@@ -0,0 +1,132 @@
...
+
+    return htmlSafe(contents);
+  }),
+
+  contentsForMinimumLines: computed('minimumLines', function() {
+    let minimumLines = this.get('minimumLines');
+    let content = '';
+
+    for (let line = 0; line < minimumLines; line++) {
+      content += '<br>';
```
andrewpye: I know this is borrowed from the React code, but maybe use string interpolation here?

addon/components/polaris-text-field/resizer.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    return htmlSafe(content);
+  }),
+
+  handleHeightCheck() {
+    let [ contentNode, minimumLinesNode ] = this.element.querySelectorAll('.Polaris-TextField__DummyInput');
```
andrewpye: Smart 👍

addon/components/polaris-text-field/resizer.js
```diff
@@ -0,0 +1,132 @@
...
+    let newHeight = Math.max(contentHeight, minimumHeight);
+    let { currentHeight, onHeightChange } = this.getProperties('currentHeight', 'onHeightChange');
+
+    if (newHeight !== currentHeight) {
+      onHeightChange(newHeight);
+    }
+  },
+
+  addResizeHandler() {
+    this.element.addEventListener('resize', this.handleHeightCheck.bind(this));
```
andrewpye: Use `ember-lifeline` for this event listener as described in other comments - removes the need to explicitly remove the event listener and faff around with the `bind` call 🙌

addon/templates/components/polaris-text-field.hbs
```diff
@@ -0,0 +1,118 @@
+<!-- template-lint invalid-interactive=false -->
```
andrewpye: Can this go right above the line that causes the linter to get angry?

addon/templates/components/polaris-text-field.hbs
```diff
@@ -0,0 +1,118 @@
...
+
+<div class="Polaris-Labelled__LabelWrapper">
+  <div class="Polaris-Label">
+    <label id="{{id}}Label" for={{id}} class="Polaris-Label__Text">{{label}}</label>
+  </div>
+  {{#if labelAction}}
+    {{polaris-button
+      plain=true
+      text=labelAction.content
```
andrewpye: We tend to use `text` instead of `content` owing to possible issues if someone passes an Ember proxy object (which has its own `content` property already)

addon/templates/components/polaris-text-field.hbs
```diff
@@ -0,0 +1,118 @@
...
+      >
+    {{/if}}
+
+    {{#if suffix}}
+      <div class="Polaris-TextField__Suffix" id="{{id}}Suffix">
+        {{render-content suffix}}
+      </div>
+    {{/if}}
+
+    {{#if (and (eq type "number") (not disabled))}}
```
andrewpye: Where possible we try to avoid using non-standard helpers to reduce the dependency on other addons. In this case it should be easy to define a `shouldShowSpinner` CP and use that instead 👍

addon/templates/components/polaris-text-field/connected.hbs
```diff
@@ -0,0 +1,21 @@
+{{#if (or left right)}}
+  <div class="Polaris-Connected">
+    {{#if left}}
+      <div class="Polaris-Connected__Item Polaris-Connected__Item--connection">
+        {{render-content left}}
```
andrewpye: I think this should just be `{{component left}}` instead of `render-content`, since it doesn't make sense to render text here as the styling looks wack if you do. Also the React implementation explicitly takes a `React.ReactNode` for this property, which in our Ember port usually translates to a component. We can still accept a string though as the `component` helper will work with that too 👍

addon/templates/components/polaris-text-field/connected.hbs
```diff
@@ -0,0 +1,21 @@
...
+
+    {{#if right}}
+      <div class="Polaris-Connected__Item Polaris-Connected__Item--connection">
+        {{render-content right}}
+      </div>
+    {{/if}}
+  </div>
+{{else}}
+  {{yield}}
+{{/if}}
```
andrewpye: Can you move this out into a separate `polaris-connected` component please, as per the React library? If you feel like it you could even do it under a separate PR and rebase this branch onto it 😛

addon/templates/components/polaris-text-field/resizer.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="Polaris-TextField__DummyInput">
+  {{finalContents}}
+</div>
+<div class="Polaris-TextField__DummyInput">
```
andrewpye: This shouldn't be rendered unless there's a `minimumLines` set

addon/templates/components/polaris-text-field/spinner.hbs
```diff
@@ -0,0 +1,26 @@
+<!-- template-lint invalid-interactive=false -->
```
andrewpye: Move this above the line that causes the linter error for explicitness

addon/templates/components/polaris-text-field/spinner.hbs
```diff
@@ -0,0 +1,26 @@
...
+
+<div
+  class="Polaris-TextField__Spinner"
+  aria-hidden="true"
+  onclick={{action onClick}}
```
andrewpye: Does this work as expected? I'd imagine it'll pass an event object as the first param, is that something that we should avoid?

tests/integration/components/polaris-text-field-test.js
```diff
@@ -0,0 +1,222 @@
...
+    assert.equal(findAll(textFieldMultilineSelector).length, 1, 'it renders a textarea');
+  });
+
+  test('it renders an actionable link when a `labelAction` hash is present', async function(assert) {
+    this.setProperties({
+      labelActionText: label,
+      labelActionCalled: false
+    });
+
+    this.set('labelAction', () => {
```
andrewpye: Any reason this isn't in the `setProperties` call above?

tests/integration/components/polaris-text-field-test.js
```diff
@@ -0,0 +1,222 @@
...
+    assert.equal(prefixNode.textContent.trim(), prefix, 'prefix value is correct');
+
+    let suffixNode = find(textFieldSuffixSelector);
+    assert.ok(suffixNode, 'it renders the suffix element');
+    assert.equal(suffixNode.textContent.trim(), suffix, 'suffix value is correct');
+  });
+
+  test('it renders as a connected component when `connectedLeft` or `connectedRight` is present', async function(assert) {
+    this.setProperties({
+      left,
```
andrewpye: These should be components :wink:

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
+import Component from '@ember/component';
+import layout from '../templates/components/polaris-text-field';
+import { createUniqueIDFactory } from '../utils/other';
```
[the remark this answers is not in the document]
tomnez: I'm good with removing it if you are, I was just matching Polaris as closely as possible. `guidFor` will give us some crazy long IDs, whereas the above gives us `TextField1, TextField2`, etc.

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+import { assert } from '@ember/debug';
+import { computed } from '@ember/object';
+import { htmlSafe } from '@ember/string';
+import { debounceTask, runDisposables } from 'ember-lifeline';
```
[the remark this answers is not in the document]
tomnez: Ah excellent I'll give that try

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+
+    id = id || getUniqueID();
+
+    this.setProperties({
+      height: null,
+      focus: false,
+      id
+    });
+
+    next(this, ()=> {
```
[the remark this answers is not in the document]
tomnez: I think I was having some issues but let me try again.

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+
+  /**
+   * Visually hide the label
+   *
+   * @property labelHidden
+   * @public
+   * @type {Boolean}
+   * @default null
+   */
+  labelHidden: null,
```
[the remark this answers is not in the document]
tomnez: 😞 ahh

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+   * @default null
+   */
+  connectedRight: null,
+
+  /**
+   * An element connected to the left of the input
+   *
+   * @property connectedLeft
+   * @public
+   * @type {String|Component}
```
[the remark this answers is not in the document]
tomnez: I'm ok with this change. Polaris _does_ let you pass just a plain string and renders it the same way shown in my screenshots, but if you're ok with not having that functionality then I'm good with just passing Component types.

addon/templates/components/polaris-text-field.hbs
```diff
@@ -0,0 +1,116 @@
...
+    onclick={{action "handleClick"}}
+  >
+
+    {{#if prefix}}
+      <div class="Polaris-TextField__Prefix" id="{{id}}Prefix">
+        {{render-content prefix}}
+      </div>
+    {{/if}}
+
+    {{#if multiline}}
```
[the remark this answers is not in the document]
tomnez: These two have slightly different attributes though, (for example `input` can take `max` and `min` whereas `textarea` doesn't). Should we use the wrapper and just let browsers do their thing when it comes to attributes that don't belong?

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+   *
+   * @property labelAction
+   * @public
+   * @type {Object}
+   * @default null
+   *
+   * Currently supports:
+   * { onAction, content, accessibilityLabel }
+   */
+  labelAction: null,
```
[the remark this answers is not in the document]
tomnez: Hm no I can move it, I must have added this later on in development.

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+   * @property onBlur
+   * @public
+   * @type {Function}
+   * @default noop
+   */
+  onBlur() {},
+
+  addValueListener() {
+    let [ field ] = this.element.querySelectorAll('input, textarea');
+    let _this = this;
```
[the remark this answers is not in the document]
tomnez: The context is different when passing methods to event listeners as callbacks though. If I were to use `this` in my arrow function callback, the "this" context refers to the html node, and not the Ember object.

addon/components/polaris-text-field.js
tomnez: You'd still get the `Polaris-TextField` class though 😄 

I'm good with `.join(' ' )`, I'll update it.

addon/components/polaris-text-field.js
tomnez: Ah no, that's an artifact from before I switched over to using regular HTML inputs. Good eyes 👀

addon/components/polaris-text-field.js
tomnez: Works either way, I can update it.

addon/components/polaris-text-field.js
tomnez: Ahh more artifacts - yes I need to update the way the input is grabbed here.

addon/components/polaris-text-field.js
tomnez: Hm yeah it would. Weird, I'll get rid of the `String` call.

addon/components/polaris-text-field.js
tomnez: Do we need to do that? `onChange` is defaulted to a no-op, so this should be safe to just call at any time.

addon/components/polaris-text-field/resizer.js
tomnez: This value doesn't seem to ever change so I don't think it requires binding, it just needs an initial value. I'll update to `ariaHidden`.

addon/components/polaris-text-field/resizer.js
tomnez: Nope! But I feel like the Andy linter does, so if you'd like I can shift things onto one line, although we risk having it look different than the React code and harder to pick out differences.

addon/templates/components/polaris-text-field.hbs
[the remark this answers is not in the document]
andrewpye: True. You could do something like create a curried component definition and use that, something like this:

```hbs
{{#with (component "wrapper-element" name=name <put all the common attributes here> as |inputElement|}}
  {{#if multiline}}
     {{component inputElement tagName="textarea" ...}}
  {{else}}
     {{component inputElement tagName="input" max=max min=min ...}}
  {{/if}}
{{/with}}
```

Not sure if that's just clunky though, what do you reckon? Definitely DRYs up the code, but maybe at the expense of readability?

addon/components/polaris-text-field.js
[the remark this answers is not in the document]
andrewpye: The `ember-lifeline` `this.addEventListener` method should take care of this for you 😉

addon/templates/components/polaris-text-field/spinner.hbs
tomnez: Seems harmless but we can remove `onclick` if you want. I got into the habit of doing this so the action would get called first, but in this case it doesn't really matter so I can update it if you want.

addon/components/polaris-text-field.js
andrewpye: I'm good with it staying there since it's come from the React code, up to you 🤷‍♂️

tests/integration/components/polaris-text-field-test.js
tomnez: Ah no, added it later on and missed this. Will move.

addon/components/polaris-text-field.js
andrewpye: Ah true, OK fine 😄

tests/integration/components/polaris-text-field-test.js
tomnez: Well _now_ they should be, after making your updates 😛

addon/components/polaris-text-field/resizer.js
andrewpye: Yeah, I wasn't thinking about the value changing, just that it's something I've done elsewhere to declare something like an `ariaHidden` property and apply it to the component's root element using `attributeBindings` 😉

addon/components/polaris-text-field/resizer.js
andrewpye: Nah I don't mind, just surprised is all 😛

addon/templates/components/polaris-text-field.hbs
andrewpye: ESLint lets you specify to ignore a rule just on the next line - if `template-lint` will let you do that, I'd prefer it 👍

addon/templates/components/polaris-text-field/spinner.hbs
andrewpye: Don't mind too strongly either way, as long as the extra event param isn't visible anywhere 👍

addon/templates/components/polaris-text-field.hbs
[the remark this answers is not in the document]
tomnez: If we have to clean this up then I'm good with the above, it _is_ DRY but yeah might be weird to read. I was originally going for something similar (hence my leftover `input` computed property) but ended up scrapping it. This seems ok though 👍

tests/integration/components/polaris-text-field-test.js
tomnez: Just leaving this comment as an update to the above comments, we decided to keep the left and right connected attributes to accept both String or Component types to match React.

tomnez:
@andrewpye updated. Some things to note about the updates -

- Left the `connectedLeft` and `connectedRight` tests as-is since we decided to keep it the same as React and accept a String or Component.
- `attributeBindings` on the `wrapper-element` were throwing an error regarding adding `attributeBindings` to a tagless component, so I left as-is.
- Computing `fieldElement` property wasn't working because it tries to compute before the DOM is available, so rather than using a CP I have it being set inside the `didInsertElement` hook to ensure that the field is available.
- Holy crap `ember-lifeline` is awesome.

addon/templates/components/polaris-text-field/spinner.hbs
andrewpye: OK cool, thanks for checking 🙂

andrewpye:
@tomnez few comments before I go over the code again:

- if this is now using #149 to bring in `polaris-connected`, can you change the base for the PR to that branch instead of `master` please?

- re. broken attribute binding on `wrapper-element`, check out #150 which should fix it. If it's still useful to you then we can merge it and you can rebase to bring it in, otherwise I think it's worth having anyway to avoid being bitten by the same gotcha in the future 😉 

- you've got a couple of failing tests which need to be addressed 👍

andrewpye:
also re. `fieldElement` as a computed property, sounds like you were just accessing it too early somehow (i.e. before the component had rendered). Probably something that could be sorted out using the run loop but no biggie 🤷‍♂️

addon/components/polaris-text-field.js
```diff
@@ -392,79 +389,82 @@ export default Component.extend({
...
-    this._super(...arguments);
-
-    let { input, focused } = this.getProperties('input', 'focused');
+  checkFocus() {
+    let { fieldElement, focused } = this.getProperties('fieldElement', 'focused');
 
-    if (input && focused) {
-      input.focus();
+    if (fieldElement && focused) {
+      this.$(fieldElement).focus();
```
andrewpye: Any reason for the JQuery wrapper here? I'd prefer to not use it unless we have to - the ideal would be to be able to remove it from the addon altogether at some point...

addon/components/polaris-text-field.js
```diff
@@ -0,0 +1,520 @@
...
+const dpl = (num) => {
+  (num.toString().split('.')[1] || []).length;
+};
+
+/**
+ * Polaris text-field component.
+ * See https://polaris.shopify.com/components/forms/text-field
+ */
+export default Component.extend({
+  classNameBindings: ['labelHidden:Polaris-Labelled--hidden'],
```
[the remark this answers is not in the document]
andrewpye: #148 is now merged down to master, so you should be able to do some rebasing and refactor this to use `polaris-labelled` 👍

addon/components/polaris-text-field.js
```diff
@@ -392,79 +389,82 @@ export default Component.extend({
...
+  }).readOnly(),
+
+  shouldShowSpinner: computed('disabled', 'type', function() {
+    let { type, disabled } = this.getProperties('type', 'disabled');
+
+    return (typeOf(type) === 'number') && !disabled;
+  }).readOnly(),
+
+  addValueListener() {
+    this.addEventListener('keyup', () => {
```
andrewpye: Can you check that adding an event handler using an arrow function works in terms of tearing the handler down? I'd think `ember-lifeline` is intelligent enough to handle it properly but worth checking given that this is a leaky pattern for regular adding/removing event handlers...

addon/components/polaris-text-field.js
```diff
@@ -392,79 +389,82 @@ export default Component.extend({
...
-    if (input && focused) {
-      input.focus();
+    if (fieldElement && focused) {
+      this.$(fieldElement).focus();
     }
   },
 
-  willDestroyElement() {
-    this.removeValueListener();
+  didInsertElement() {
```
andrewpye: Call `this._super(...arguments)` here 😉

addon/components/polaris-text-field.js
```diff
@@ -392,79 +389,82 @@ export default Component.extend({
...
+    this.addValueListener();
+    this.checkFocus();
   },
 
-  destroy() {
-    runDisposables(this);
-
+  didReceiveAttrs() {
     this._super(...arguments);
+    this.checkFocus();
```
andrewpye: This might need to be scheduled to use the run loop, since I'm not sure the attributes will have updated by the time this runs (you're probably alright though)

addon/components/polaris-text-field/resizer.js
```diff
@@ -13,14 +14,15 @@ const ENTITIES_TO_REPLACE = {
...
-};
+}
 
-export default Component.extend({
+export default Component.extend(ContextBoundTasksMixin, ContextBoundEventListenersMixin, {
   classNames: ['Polaris-TextField__Resizer'],
+  attributeBindings: ['ariaHidden:aria-hidden'],
 
-  'aria-hidden': true,
+  ariaHidden: "true",
```
andrewpye: Single quotes in JS files pls 😉

addon/templates/components/polaris-text-field.hbs
```diff
@@ -1,13 +1,11 @@
-<!-- template-lint invalid-interactive=false -->
-
 <div class="Polaris-Labelled__LabelWrapper">
```
andrewpye: Rebase onto `master` to pull in `polaris-labelled` and refactor all this, it should be much more straightforward then 😉

addon/components/polaris-text-field.js
tomnez: Ahh no I can use `fieldElement.focus()` 🤦‍♂️

addon/components/polaris-text-field.js
tomnez: I think we're safe, the example from their docs shows:

```javascript
didInsertElement() {
    this._super(...arguments);
    addEventListener(this, window, 'scroll', (e) => {
      this.set('windowScrollOffset', e.clientY);
    });
  },
```

The only difference is that I'm using the Mixin to create mine which uses the same `addEventListener` method above.

addon/components/polaris-text-field.js
tomnez: Did a quick test, we good -

focus

addon/components/polaris-text-field/resizer.js
```diff
@@ -0,0 +1,128 @@
...
+      return;
+    }
+
+    let contentHeight = contentNode.offsetHeight;
+    let minimumHeight = minimumLinesNode.offsetHeight;
+    let newHeight = Math.max(contentHeight, minimumHeight);
+    let { currentHeight, onHeightChange } = this.getProperties('currentHeight', 'onHeightChange');
+
+    if (newHeight !== currentHeight) {
+      next(this, () => {
```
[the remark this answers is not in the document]
andrewpye: I'd imagine the flow is something like this:

- `resizer` calls `onHeightChange`
- parent `text-field` component handles `onHeightChange`, which updates the `height` property
- parent `text-field` component passes updated `height` property into `resizer` component
- `resizer` component's `didUpdateAttrs` hook ends up invoking `onHeightChange` again because `currentHeight` hasn't updated yet, taking us around the loop a second time...

So the way to handle this would be to stop `onHeightChange` running more than once (especially in a single run loop). For that reason, `scheduleOnce` is probably more appropriate than `next` I'd say...

tomnez:
@andrewpye updated and added docs.

[Pretty rendered docs](https://github.com/smile-io/ember-polaris/blob/polaris-text-field/docs/text-field.md#text-field) here.

PR #147, opened 2018-07-02
Allow passing component name and props hash to data table cells

andrewpye:
Passing dynamic, data-driven components to `polaris-data-table` cells is currently _very_ difficult owing to the impossibility of generating components for rendering outside template files. This PR aims to solve that issue by adding the ability to pass a hash with a `componentName` and optional `props` hash as cell data - the table will then render that component with the provided properties. This makes it possible to process e.g. an array of raw cell data in JS-land to instruct the table to render components in particular cells.


andrewpye:
>Obviously we should look if we can include the template compiler with the addon
Would be nice, but I'd prefer to avoid adding it unless it's needed... not sure how to handle this really at the moment, if you have any suggestions I'd appreciate it!

>Does the layout really need to be a CP? Wondering if this will be subject to double rendering issues
I guess it's either a CP or gets explicitly set in `didUpdateAttrs`... didn't see any issues testing this but can push it harder to see if it breaks.

>Will this have issues accessing the Ember registry? Like if there's a service injected in the component that's passed in, will that work?
Not entirely sure what you mean... this renders the "proxied" component the same way it'd get rendered by any other template, as far as I can tell. Can also test this before going any further.

vladucu:
> Would be nice, but I'd prefer to avoid adding it unless it's needed... not sure how to handle this really at the moment, if you have any suggestions I'd appreciate it!

well, if we add this, would be a requirement no? yeah...will try to look into it a bit too

> Not entirely sure what you mean... this renders the "proxied" component the same way it'd get rendered by any other template, as far as I can tell. Can also test this before going any further.

yeah, actually, I think you are right....but yeah, let's test in an app, to be sure

andrewpye:
>>Would be nice, but I'd prefer to avoid adding it unless it's needed... not sure how to handle this really at the moment, if you have any suggestions I'd appreciate it!

>well, if we add this, would be a requirement no? yeah...will try to look into it a bit too

I think it'd only error if a host app tries to use this new functionality - everything else should still be fine 🤞

PR #148, opened 2018-07-02
Add label and labelled components

andrewpye:
Adds the internal `polaris-label` and `polaris-labelled` components, which are needed to build the other form controls like `polaris-text-field`, `polaris-range-slider` etc.

### DIFF
--- a/addon/components/polaris-label.js
+++ b/addon/components/polaris-label.js
@@ -1,0 +1,53 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import layout from '../templates/components/polaris-label';
+
+/**
+ * Internal Polaris label component.
+ */
+export default Component.extend({
+  // Tagless component so that Ember doesn't apply the `id`
+  // attribute to the component's root element.
+  tagName: '',
+
+  layout,
+
+  /**
+   * Label content
+   *
+   * This component can be used in block form,
+   * in which case the block content will be used
+   * instead of `text`
+   *
+   * @type {String|Component}
+   * @public
+   */
+  text: null,
+
+  /**
+   * A unique identifier for the label
+   *
+   * @type {String}
+   * @public
+   */
+  id: null,
+
+  /**
+   * Visually hide the label
+   *
+   * @type {Boolean}
+   * @default false
+   * @public
+   */
+  hidden: false,
+
+  /**
+   * ID for the label element
+   *
+   * @type {String}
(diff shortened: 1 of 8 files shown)


addon/components/polaris-label.js
```diff
@@ -0,0 +1,53 @@
...
+   */
+  hidden: false,
+
+  /**
+   * ID for the label element
+   *
+   * @type {String}
+   * @private
+   */
+  labelID: computed('id', function() {
```
tomnez: nit: `Id` rather than `ID` (seems much more widely used in this addon)

addon/components/polaris-labelled.js
```diff
@@ -0,0 +1,100 @@
...
+    return guidFor(this);
+  }),
+
+  /**
+   * ID for the error message div
+   *
+   * @type {String}
+   * @private
+   */
+  errorID: computed('id', function() {
```
tomnez: Same for these regarding `ID`, but also since you have a few of these in here I'm not particularly stuck to this if you think it doesn't matter.

addon/components/polaris-label.js
andrewpye: Yeah, was matching the React implementation but having `ID` bugs me, updating 😛

addon/templates/components/polaris-label.hbs
```diff
@@ -0,0 +1,9 @@
+<div class="Polaris-Label">
```
vladucu: missing some [hidden class](https://github.com/Shopify/polaris/blob/master/src/components/Label/Label.tsx#L20) here

addon/templates/components/polaris-labelled.hbs
```diff
@@ -0,0 +1,37 @@
+<div class={{if labelHidden "Polaris-Labelled--hidden"}}>
+  {{#if label}}
+    <div class="Polaris-Labelled__LabelWrapper">
+      {{#polaris-label id=id hidden=false}}
```
vladucu: totally fine, could have just passed `text` here

andrewpye: True, was just following the React component...

addon/templates/components/polaris-label.hbs
andrewpye: @vladucu good catch, thanks! Fixed in #151 👍

PR #149, opened 2018-07-02
Polaris connected

tomnez:
### Overview

Implements the [polaris-connected](https://github.com/Shopify/polaris/tree/master/src/components/Connected) component.

@andrewpye this was taken straight from my `polaris-text-field` branch. It seems the React implementation also includes some `onFocus` and `onBlur` events for the connected items. Do we want to refactor and add those as well? Also it seems that this component isn't documented in the Polaris docs, should we add docs or should we follow suit and use this in the same manner as more of an undocumented component used for building other components?

### Demo

This isn't the most beautiful example, but it displays passing either a `String` or `Component` to the connected sides:

template:
```hbs
{{#polaris-connected
    left="as a string"
    right=(component "polaris-button"
      text="As a component"
      onClick=(action "handleClick")
    )
  }}
    <input class="Polaris-TextField__Input" type="text" value="Hello">
  {{/polaris-connected}}
```

result:
screen shot 2018-07-02 at 4 04 42 pm


tomnez:
[Implement polaris-connected component](https://trello.com/c/Y5VZBwdQ/164-implement-polaris-connected-component)

tests/integration/components/polaris-connected-test.js
```diff
@@ -0,0 +1,91 @@
...
+        </div>
+      {{/polaris-connected}}
+    `);
+
+    const connectedComponentNode = find(connectedContainerSelector);
+    const leftContentNode = find(leftContentSelector);
+    const rightContentNode = find(rightContentSelector);
+    const primaryContent = find('.primary-content');
+
+    assert.notOk(connectedComponentNode, 'connected component container does not get rendered');
```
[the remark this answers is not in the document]
vladucu: yeah, I find that much nicer and better error messages....also, seems like it might be baked in ember-cli in the near future

tests/integration/components/polaris-connected-test.js
```diff
@@ -0,0 +1,83 @@
...
+        </div>
+      {{/polaris-connected}}
+    `);
+
+    const primaryContentSelector = '.primary-content';
+
+    assert.dom(connectedContainerSelector).doesNotExist('connected component container does not get rendered');
+    assert.dom(leftContentSelector).doesNotExist('left content node does not get rendered');
+    assert.dom(rightContentSelector).doesNotExist('right content node does not get rendered');
+    assert.dom(primaryContentSelector).exists('the block content still renders when connected sides are not passed in');
```
vladucu: nit: this `exists` test is basically also tested by the `hasText` below (same below)

tomnez: Hm I'd agree if the `primaryContentSelector` wasn't nested inside a conditional in the component template. In this case there's both a chance that the element may or may not get rendered, as well as the chance that the value inside the element could be wrong.

Only testing `hasText` and having it fail might lead you to believe the element itself never got rendered, when it could just be that the value got wrongly mutated somehow. Although now I'm also just being nitpicky because I'm sure the error message would be pretty explicit 😝

vladucu: yeah, no worries...fine with this.....but yeah, that's what I'm ...this will catch your error whatever that is, cause you can't check a text if the element doesn't exist

PR #150, opened 2018-07-03
Auto-bind wrapper-element attributes

andrewpye:
Allows us to pass arbitrary attributes to the internal `wrapper-element` component and have them be bound to the rendered element (if there is one).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # ember-polaris Changelog
 
 ### Unreleased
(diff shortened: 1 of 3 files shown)


andrewpye:
I don't think it's particularly hacky, it just dynamically generates `attributeBindings`. This is basically how [`ember-one-way-controls`](https://github.com/DockYard/ember-one-way-controls/blob/master/addon/-private/dynamic-attribute-bindings.js), and was originally suggested by Robert Jackson himself (can't find the post now)... I think I'd prefer to do this than pull in another addon, though open to discussion on it

PR #151, opened 2018-07-03
Add missing hidden class to `polaris-label`

andrewpye:
Adds the `Polaris-Label--hidden` class that was missing from `polaris-label` when `hidden` was truthy.


vladucu:
PS: this might worth a test

PR #152, opened 2018-07-03
Accept card title as a string or a component

vladucu:


tests/integration/helpers/is-component-definition-test.js
```diff
@@ -0,0 +1,17 @@
...
+module('Integration | Helper | is-component-definition', function(hooks) {
+  setupRenderingTest(hooks);
+
+  // Replace this with your real tests.
+  test('it renders', async function(assert) {
+    this.set('inputValue', '1234');
+
+    await render(hbs`{{is-component-definition inputValue}}`);
+
+    assert.equal(this.element.textContent.trim(), '1234');
```
andrewpye: This test needs updating, surely it doesn't pass?

addon/helpers/is-component-definition.js
```diff
@@ -0,0 +1,13 @@
+import { helper } from '@ember/component/helper';
+import { typeOf } from '@ember/utils';
+
+export function isComponentDefinition([ content ]) {
+  if (typeOf(content) !== 'object') {
+    return false;
+  }
+
+  let contentConstructorName = content.constructor.name || '';
+  return contentConstructorName.indexOf('ComponentDefinition') > -1;
```
andrewpye: Should probably update `render-content` to use this function to keep things DRY 👍

addon/templates/components/polaris-card.hbs
```diff
@@ -3,7 +3,11 @@
     <div class="Polaris-Card__Header">
       {{#polaris-stack alignment="baseline" as |stack|}}
         {{#stack.item fill=true}}
-          {{polaris-heading text=title}}
+          {{#if (is-component-definition title)}}
```
andrewpye: Hmm... this helper is a nice idea, but it will ignore the changes I made to `render-content` in #147 (handling `content` being a hash of `componentName` and `props`). We probably want to be able to support both the component definition and `componentName` + `props` hash way of doing things here, since `render-content` will support it...

addon/components/polaris-card.js
```diff
@@ -62,6 +63,7 @@ export default Component.extend({
    * @property headerActions
    * @type {Action[]}
```
andrewpye: This should probably get changed to `Object[]` now, maybe worth checking if any other `@type`s need updating too

tests/integration/helpers/is-component-definition-test.js
vladucu: many tests should be updated if we want to be consistent with latest ones...though this should work well (ember always has backwards support)

addon/helpers/is-component-definition.js
vladucu: yeah, can do

addon/templates/components/polaris-card.hbs
vladucu: not really an idea...we need to not wrap title in a heading if it's passed as a component....but will check if we can support both

addon/components/polaris-card.js
vladucu: will leave for when we cleanup/refactor...just cause there are many things we'd need to update here

vladucu:
@andrewpye updated, dry-ed up the `render-content` component
since the `componentName` is not yet merged, will leave that part for once we settled on the solution for that, if that's cool with you

PR #153, opened 2018-07-03
Add indicator component

vladucu:


vladucu:
closing this, component is in the button component but not actually put to use

PR #156, opened 2018-07-03
Add prettier

vladucu:
Configures prettier, all cool kids seem to do it now (*cough* Shopify), mostly putting this up to see what's the general feeling.

For now you can run `yarn lint:js --fix` to format, if we'll want to go forward we can maybe have precommits too, but there's good support in editors too


tomnez:
I like this, although I personally would prefer breaking up arrays onto single line indices even if the line width was under `80` characters (or whatever value we go with). I assume that's configurable if we all agree on it?

That's really the only nitpicky opinion I have regarding using this. All the other rules seem reasonable and it will be nice to have something to catch common style errors before making a PR 👍

addon/-private/drop-zone-state.js
```diff
@@ -44,11 +44,11 @@ export default EmberObject.extend({
 
   overlayText: computed('type', function() {
     let type = this.get('type');
-    return get(dropZone, `overlayText${ capitalize(type) }`);
+    return get(dropZone, `overlayText${capitalize(type)}`);
```
andrewpye: Can (should?) we configure to keep these spaces?

addon/components/polaris-avatar.js
```diff
@@ -22,29 +18,16 @@ const avatarImages = [
   'avatarSix',
   'avatarSeven',
   'avatarEight',
-  'avatarNine'
+  'avatarNine',
```
andrewpye: This doesn't add an extra empty item to the array? TIL :smile:

andrewpye:
@vladucu looks like build failures here, needs fixing before merging 🙏

addon/-private/drop-zone-state.js
vladucu: will look into this if possible

vladucu:
thx for the feedback

will look into the things pointed out (ex: breaking up arrays onto single line indices even if the line width was under 80 characters)....and after that on the build failures if we want to go with this

PR #158, opened 2018-07-09
Pass selector to value listener in polaris-text-field

tomnez:
### Overview

The `polaris-text-field` uses ember-lifeline to add an event listener for the `keyup` event on the input field. It [doesn't require](https://github.com/ember-lifeline/ember-lifeline/blob/master/addon/mixins/dom.js#L57) a selector to be passed in, and worked fine in the dummy app. However when pulled into other apps it seems to be having issues (throws error: `Called addEventListener with selector not found in DOM: keyup`) so to be safe I'm passing the selector, which fixes the error and still works as expected.

### DIFF
--- a/addon/components/polaris-text-field.js
+++ b/addon/components/polaris-text-field.js
@@ -422,7 +422,7 @@ export default Component.extend(ContextBoundTasksMixin, ContextBoundEventListene
   }).readOnly(),
 
   addValueListener() {
-    this.addEventListener('keyup', () => {
+    this.addEventListener('input, textarea', 'keyup', () => {
       this.debounceTask('debouncedUpdateValue', 250);
     });
   },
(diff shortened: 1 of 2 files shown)


addon/components/polaris-text-field.js
```diff
@@ -422,7 +422,7 @@ export default Component.extend(ContextBoundTasksMixin, ContextBoundEventListene
   }).readOnly(),
 
   addValueListener() {
-    this.addEventListener('keyup', () => {
+    this.addEventListener('input, textarea', 'keyup', () => {
```
vladucu: should we prefer `this.element.querySelector('input')`....not sure if having just `input` will add listener to all inputs on the page or not

tomnez: @vladucu It uses `findElement(this.element, selector)`, so I think it's scoped to `this.element` of the component.

https://github.com/ember-lifeline/ember-lifeline/blob/master/addon/mixins/dom.js#L105

vladucu: sweet, thx for checking

PR #159, opened 2018-07-10
Add notes about avoiding Glimmer error under dynamic usages of auto-wrapping components

andrewpye:
At present, 4 of the components in `ember-polaris` auto-wrap their child elements in `div`s - `polaris-button-group`, `polaris-form-layout`, `polaris-form-layout/group` and `polaris-stack`. If any of these components are used with dynamic contents (e.g. with an `if`- or `each`-block inside), whenever an item is added or removed a Glimmer error can occur. Unfortunately I don't think we can detect this possibility and avoid it automatically (certainly not quickly), so I thought it was worth adding notes and examples to the docs for those components affected.

### DIFF
--- a/docs/form-layout.md
+++ b/docs/form-layout.md
@@ -28,3 +28,23 @@ Form layout with two groups, the second of which is condensed:
   {{/formLayout.group}}
 {{/polaris-form-layout}}
 ```
+
+For dynamic forms where items are added or removed once the form has rendered, you must explicitly a form layout item component around each item to prevent a Glimmer error such as `Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node`:
+
+```hbs
+{{#polaris-form-layout as |formLayout|}}
+  {{#each formItems as |formItem|}}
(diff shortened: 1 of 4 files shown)


docs/form-layout.md
```diff
@@ -28,3 +28,23 @@ Form layout with two groups, the second of which is condensed:
   {{/formLayout.group}}
 {{/polaris-form-layout}}
 ```
+
+For dynamic forms where items are added or removed once the form has rendered, you must explicitly a form layout item component around each item to prevent a Glimmer error such as `Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node`:
```
tomnez: typo: you have "explicitly a form item" instead of "explicitly _add_ a form item"

andrewpye: 😱 

Good catch, thanks 😄

andrewpye:
Gonna merge this down without making a release, since there are no code changes to go out and people probably read the docs online on here rather than from their app's `node_modules` directory 😉

PR #160, opened 2018-07-26
Fixes svg loading error in `polaris-drop-zone`

### LINKED ISSUE #23: Look into proper import of icon SVG files
The initial implementation of the `polaris-icon` component has a few nasties around SVG import, such as having to add `svg` options to the host app's `ember-cli-build.js`. We should look into how we do this properly.

sivakumar-kailasam:


andrewpye:
@sivakumar-kailasam this would be awesome to get sorted! Currently host apps have to manually copy the SVGs from the React component library into their app's assets... do you think this is something that could be automated here as well?

andrewpye:
If we can avoid having to pollute the host app with the SVGs that'd be pretty cool 🙂 We'd need to be careful though because a bunch of the icons come from the React implementation's `components` subdirectories (e.g. the [`Avatar` component](https://github.com/Shopify/polaris/tree/master/src/components/Avatar/images)). The current usage of this addon needs those icons dumped into the same directory as the [main icon collection](https://github.com/Shopify/polaris/tree/master/src/icons) (yes, it's ugly 😅) so there'd likely be some breakages in existing apps and possibly some of the components in this addon...

sivakumar-kailasam:
I took at the compiled `@shopify/polaris` module and looks like we'll have resort to file parsing instead of requiring the file because ember-cli freaks out at the dependencies of that module. Even ember-auto-import couldn't help here.

So I'm going to file an issue in polaris asking if they can export it as a separate module in their builds & lets tackle that in a different PR

index.js
```diff
@@ -4,6 +4,7 @@ const path = require('path');
 const resolve = require('resolve');
 const Funnel = require('broccoli-funnel');
 const MergeTrees = require('broccoli-merge-trees');
+// const polaris = require('@shopify/polaris')
```
andrewpye: Drop this if it's not being used 😉

tests/integration/components/polaris-drop-zone-test.js
```diff
@@ -241,17 +241,13 @@ module('Integration | Component | polaris-drop-zone', function(hooks) {
         assert.dom(fileUploadStackSelector).exists('fileUpload wraps content in a stack');
         assert.dom(fileUploadStackSelector).hasClass('Polaris-Stack--vertical', 'fileUpload stack is vertically aligned');
         assert.dom(fileUploadImageSelector).exists('fileUpload image is rendered');
-        assert.dom(fileUploadImageSelector).hasAttribute('src', '/ember-polaris/images/file-upload.svg', 'fileUpload has correct image `src` attribute');
```
andrewpye: Can you replace these tests with tests that the correct SVG renders please? We have a mock `svg-jar` component which sets the source on the element's data set which we use e.g. [here](https://github.com/smile-io/ember-polaris/blob/master/tests/integration/components/polaris-action-list-test.js#L106)

addon/templates/components/polaris-drop-zone/file-upload.hbs
```diff
@@ -13,11 +13,11 @@
 {{else if (eq size "large")}}
   {{#polaris-stack vertical=true spacing="tight"}}
     {{#if (eq type "file")}}
-      <img class={{imageClasses}} alt="" src={{assetFileUpload}}>
+      {{svg-jar assetFileUpload class=imageClasses}}
```
vladucu: any reason for rendering these as `svg` instead of `img`? Just asking cause this is what Polaris does and also if we can avoid `svg-jar` processing that's also a plus I think

sivakumar-kailasam: In its current form the images don't load in the consuming app because these assets aren't added to the host app's assets. This is why I started looking into this 😄 

I looked at removing svg-jar before making this change but it looks like we do it use it in other components within this app, so went with this.

sivakumar-kailasam:
> update the readme to reflect the changes to the installation procedure.
@andrewpye looks like this will have to remain as is for now until we sort out the issue with importing polaris icons.

andrewpye:
>>update the readme to reflect the changes to the installation procedure.

>@andrewpye looks like this will have to remain as is for now until we sort out the issue with importing polaris icons.

@sivakumar-kailasam all of it? Maybe I've misunderstood, I thought this PR removed the need for adding the `ember-svg-jar` options to the host app's `ember-cli-build.js` - did I get that wrong? 😅

sivakumar-kailasam:
You did mean https://github.com/smile-io/ember-polaris/blob/master/README.md#icons section right?  

This PR only helps polaris-drop-zone 😄 other assets still need to be imported as mentioned in that readme

andrewpye:
Ahhh OK, from the PR title and the comment about this fixing #23 I thought this was something much wider-scoped than it actually is 😅 In that case can you update the title to reflect that this only impacts `polaris-drop-zone`, and remove the "github magic" about this fixing #23 because it doesn't actually address what that issue was opened for? Thanks!

addon/templates/components/polaris-drop-zone/file-upload.hbs
vladucu: ahh, got it
I know this worked initially, but then moved this repo to `@smile-io` org, so that broke this
I think if we change `/ember-polaris/images/file-upload.svg` -> `/@smile-io/ember-polaris/images/file-upload.svg` should work....cause the addon's assets in `public` should be merged in the host app right?

sivakumar-kailasam: Its now at `@smile-io/ember-polaris/ember-polaris/images/file-upload.svg` 😞

vladucu: damn, ok....let's just use this branch for now

sivakumar-kailasam:
Done!

PR #165, opened 2018-08-29
Add is developing addon based on flag

vladucu:
Adding this to ease development/debugging

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,10 @@ const MergeTrees = require('broccoli-merge-trees');
 module.exports = {
   name: '@smile-io/ember-polaris',
 
+  isDevelopingAddon() {
+    return process.env.SMILE_DEV;
+  },
+
   treeForStyles(tree) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -8,6 +8,10 @@ const MergeTrees = require('broccoli-merge-trees');
 module.exports = {
   name: '@smile-io/ember-polaris',
 
+  isDevelopingAddon() {
+    return process.env.SMILE_DEV;
```
andrewpye: Do we want something more general than `SMILE_DEV` on this? Or are we only allowing Smile team to develop this addon? 😅

vladucu: I think the idea is to be something more unique to avoid anyone having this by mistake and have this result in files being watched and increase build times

vladucu: of course, anyone can set this flag if their developing 😄

andrewpye: Yeah I know 😛

PR #166, opened 2018-09-11
Safeguard against non finite numbers in progress bar

tomnez:
Prevents any non-finite number being used as a progress percentage in the `polaris-progress-bar`.

Alternatively this could also be caught in a `didReceiveAttrs` hook, but that might be overkill since the `progress` attr seems to only be used in computing the `parsedProgress` property.


tomnez:
Closing this because Polaris doesn't include any checks/conditions like this.

PR #168, opened 2018-09-14
Upgrade Ember to 3.4

vladucu:
Will merge this in a main `dev` branch and basing Polaris resource list on top of these

`ember-cli` upgrade diff [here](https://github.com/ember-cli/ember-new-output/compare/v3.1.4...v3.4.2)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,18 @@
-/blueprints/*/files/**/*.js
+# unconventional js
+/blueprints/*/files/
+/vendor/
+
+# compiled output
+/dist/
+/tmp/
+
+# dependencies
+/bower_components/
+
+# misc
+/coverage/
+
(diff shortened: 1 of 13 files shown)


sivakumar-kailasam:
Looks like there's a bunch of errors with the latest ember-template-lint rules. May be turn them off for now so that tests pass?

Also looks like we're still using `ember-native-dom-helpers`.  if you run `ember-cli-update --run-codemods` there should be one which ports the imports over to 3+ ember test helpers.

vladucu:
thx guys!
yeah, didn't really looked at errors at all....will look before merging

vladucu:
shipped and following up with a PR to fixing linting...won't spend the time to convert away from native dom now, since there's manual work involved too...we'll do that as we add/update components

PR #172, opened 2018-09-18
Fix tests

vladucu:
- fixes template linting issues
- enables jquery-integration for now (until we look into removing the few usages)
- did not converted all tests to new test-helpers API from native-dom, will do that as we keep updating tweaking components or we find time, since after the codemod there's still manual work needed in quite a few places

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,15 +30,15 @@ jobs:
     - stage: "Tests"
       name: "Tests"
       script:
-        - yarn lint:hbs
-        - yarn lint:js
-        - yarn test
+        - npm run lint:hbs
+        - npm run lint:js
+        - npm test
 
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
     - stage: "Additional Tests"
       env: EMBER_TRY_SCENARIO=ember-lts-2.12
-      env: EMBER_TRY_SCENARIO=ember-lts-2.16
+    - env: EMBER_TRY_SCENARIO=ember-lts-2.16
     - env: EMBER_TRY_SCENARIO=ember-lts-2.18
     - env: EMBER_TRY_SCENARIO=ember-release
     - env: EMBER_TRY_SCENARIO=ember-beta
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ before_install:
   - export PATH=$HOME/.yarn/bin:$PATH
 
 install:
-  - yarn install --no-lockfile --non-interactive --ignore-engines
+  - yarn install --no-lockfile --non-interactive
 
 script:
   - node_modules/.bin/ember try:one $EMBER_TRY_SCENARIO
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ before_install:
   - export PATH=$HOME/.yarn/bin:$PATH
 
 install:
-  - yarn install --no-lockfile --non-interactive
+  - yarn install --no-lockfile --non-interactive --ignore-engines
 
 script:
   - node_modules/.bin/ember try:one $EMBER_TRY_SCENARIO
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - "6"
+  - "8"
 
 sudo: false
 dist: trusty
@@ -30,9 +30,9 @@ jobs:
     - stage: "Tests"
       name: "Tests"
       script:
-        - npm run lint:hbs
-        - npm run lint:js
-        - npm test
+        - yarn lint:hbs
+        - yarn lint:js
+        - yarn test
 
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
--- a/addon/templates/components/polaris-action-list/item.hbs
+++ b/addon/templates/components/polaris-action-list/item.hbs
@@ -13,7 +13,7 @@
         role="presentation"
         class="Polaris-ActionList__Image"
         style={{imageBackgroundStyle}}
-      ></div>
+      />
     {{/if}}
 
     <div class="Polaris-ActionList__Text">
--- a/addon/templates/components/polaris-action-list/item.hbs
+++ b/addon/templates/components/polaris-action-list/item.hbs
@@ -13,7 +13,8 @@
         role="presentation"
         class="Polaris-ActionList__Image"
         style={{imageBackgroundStyle}}
-      />
+      >
+      </div>
     {{/if}}
 
     <div class="Polaris-ActionList__Text">
--- a/addon/templates/components/polaris-callout-card.hbs
+++ b/addon/templates/components/polaris-callout-card.hbs
@@ -21,7 +21,7 @@
             text=primaryAction.text
             onClick=(action primaryAction.onAction)
           )
-        as |primaryButton|
+          as |primaryButton|
         }}
           {{#if secondaryAction}}
             {{#polaris-button-group}}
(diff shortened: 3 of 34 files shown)


.travis.yml
```diff
@@ -50,7 +50,7 @@ before_install:
   - export PATH=$HOME/.yarn/bin:$PATH
 
 install:
-  - yarn install --no-lockfile --non-interactive
+  - yarn install --no-lockfile --non-interactive --ignore-engines
```
vladucu: ignoring engines cause `@shopify/react-utilities` will fail our tests

.travis.yml
```diff
@@ -30,15 +30,15 @@ jobs:
     - stage: "Tests"
       name: "Tests"
       script:
-        - yarn lint:hbs
-        - yarn lint:js
-        - yarn test
+        - npm run lint:hbs
+        - npm run lint:js
+        - npm test
```
andrewpye: Interesting... guessing this change was from a codemod? Do you know what the reason for using `npm` over `yarn` is here?

addon/templates/components/polaris-callout-card.hbs
```diff
@@ -14,10 +14,14 @@
...
-          onClick=(action primaryAction.onAction))
-          as |primaryButton|
+        {{!-- template-lint-disable attribute-indentation --}}
+        {{!-- TODO remove rule above once linter handles this correctly --}}
+        {{#with
+          (component "polaris-button"
+            text=primaryAction.text
+            onClick=(action primaryAction.onAction)
+          )
+        as |primaryButton|
```
andrewpye: This looks like it should be indented IMO... any thoughts?

addon/templates/components/polaris-data-table/cell.hbs
```diff
@@ -1,8 +1,22 @@
...
   {{#if header}}
-    <th aria-hidden="true" role="presentation" id={{cellElementId}} class={{cellClassNames}} style={{style}}></th>
+    <th
+      aria-hidden="true"
+      role="presentation"
+      id={{cellElementId}}
+      class={{cellClassNames}}
+      style={{style}}
+    >
+    </th>
```
andrewpye: Are you aware of any reason for this not being made self-closing like a couple of the `div`s in this PR?

addon/templates/components/polaris-progress-bar.hbs
```diff
@@ -1,10 +1,12 @@
-<progress class="Polaris-ProgressBar__Progress"
+<progress
+  class="Polaris-ProgressBar__Progress"
   max="100"
-  value="{{parsedProgress}}"
+  value={{parsedProgress}}
```
andrewpye: Could this removal of the quotes cause issues with host apps that pass numeric values instead of strings?

.travis.yml
sivakumar-kailasam: Tried running `yarn` on this branch on local & got,
`error @shopify/react-utilities@2.0.3: The engine "node" is incompatible with this module. Expected version ">=8.9.0".`

Setting the node version of the container to `8` should solve this. Node 8 is LTS anyway.

.travis.yml
sivakumar-kailasam: the codemod doesn't account for yarn/npm usage when updating the travis files. these should be `yarn` instead of `npm run`

vladucu: oh, no...was debugging something....will revert this

addon/templates/components/polaris-callout-card.hbs
vladucu: this seems to be what the linter wants....it complains otherwise

addon/templates/components/polaris-action-list/item.hbs
```diff
@@ -13,7 +13,7 @@
         role="presentation"
         class="Polaris-ActionList__Image"
         style={{imageBackgroundStyle}}
-      ></div>
+      />
```
sivakumar-kailasam: isn't div not supposed to be a self closing tag?

addon/templates/components/polaris-data-table/cell.hbs
vladucu: no reason....just missed these

addon/templates/components/polaris-progress-bar.hbs
vladucu: I don't think they should, but good point....will double check

addon/templates/components/polaris-data-table.hbs
```diff
@@ -28,7 +35,10 @@
 
         {{#if totals}}
           <tr>
-            {{#each (polaris-data-table/insert-presentational-cell totals) as |total index|}}
+            {{#each
+              (polaris-data-table/insert-presentational-cell totals)
+            as |total index|
```
sivakumar-kailasam: doesn't this line look off?

addon/templates/components/polaris-data-table.hbs
```diff
@@ -4,13 +4,20 @@
...
+>
   <div class="Polaris-DataTable__ScrollContainer" style={{scrollContainerStyle}}>
+    {{! template-lint-disable table-groups }}
     <table class="Polaris-DataTable__Table">
       <thead>
         <tr>
-          {{#each (polaris-data-table/insert-presentational-cell headings) as |heading headingIndex|}}
+          {{#each
+            (polaris-data-table/insert-presentational-cell headings)
+          as |heading headingIndex|
```
sivakumar-kailasam: doesn't this line look off?

.travis.yml
vladucu: yeah, I tried to keep this, but I guess there's no reason we shouldn't use 8

PS there's 6 LTS too 😁

sivakumar-kailasam: ah i forget they support multiple LTS' https://github.com/nodejs/Release#release-schedule

addon/templates/components/polaris-callout-card.hbs
andrewpye: 😞 

Is that something that's configurable?

andrewpye: (not saying it needs to be done now 😉)

addon/templates/components/polaris-data-table.hbs
vladucu: I agree....but this was erroring the linter....checked again now, it doesn't...weird
will fix

addon/templates/components/polaris-action-list/item.hbs
andrewpye: Yeah I think it shouldn't be self-closing 🤷‍♂️ Does the linter complain if this is left as it was before @vladucu?

sivakumar-kailasam: lets disable the check here if it complains about it instead of making this change 🙂

addon/templates/components/polaris-callout-card.hbs
vladucu: for now, I'm gonna fix, since we disabled this rule manually above with that comment...might be a glitch with the linter (didn't checked yet if there's any related issue)

addon/templates/components/polaris-data-table/row.hbs
```diff
@@ -1,4 +1,7 @@
-{{#each (polaris-data-table/insert-presentational-cell row) as |cellText cellIndex|}}
+{{#each
+  (polaris-data-table/insert-presentational-cell row)
+as |cellText cellIndex|
```
tomnez: Missed an indent here

tests/dummy/app/templates/dropzone.hbs
```diff
@@ -1,3 +1,8 @@
-{{#polaris-drop-zone type="image" accept="image/jpeg,image/png" onDrop=(action "drop") as |dropZone|}}
+{{#polaris-drop-zone
+  type="image"
+  accept="image/jpeg,image/png"
+  onDrop=(action "drop")
+as |dropZone|
```
tomnez: Indent here too

PR #174, opened 2018-10-04
Implement Polaris inline-error component

### LINKED ISSUE #175: Implement Polaris inline-error component
Component docs [here](https://polaris.shopify.com/components/forms/inline-error)

vladucu:
fix #175

### DIFF
--- a/addon/components/polaris-inline-error.js
+++ b/addon/components/polaris-inline-error.js
@@ -1,0 +1,35 @@
+import Component from '@ember/component';
+import layout from '../templates/components/polaris-inline-error';
+import { assert } from '@ember/debug';
+
+export default Component.extend({
+  layout,
+  tagName: '',
+
+  /**
+   * Unique identifier of the invalid form field that the message describes
+   *
+   * @type {String}
+   * @default null
+   * @public
+   * @required
+   */
+  fieldID: null,
+
+  /**
+   * Content briefly explaining how to resolve the invalid form field input.
+   *
+   * @type {String}
+   * @default null
+   * @public
+   */
+  message: null,
+
+  'data-test-inline-error': true,
+  'data-test-inline-error-icon': true,
+
+  init() {
+    this._super(...arguments);
+    assert('[polaris-inline-error] Missing required `fieldID` param!', this.get('fieldID'));
+  },
+});
--- a/addon/components/polaris-inline-error.js
+++ b/addon/components/polaris-inline-error.js
@@ -19,7 +19,7 @@ export default Component.extend({
   /**
    * Content briefly explaining how to resolve the invalid form field input.
    *
-   * @type {String}
+   * @type {String|Component}
    * @default null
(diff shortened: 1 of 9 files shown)


addon/components/polaris-inline-error.js
```diff
@@ -0,0 +1,35 @@
...
+   * @public
+   */
+  message: null,
+
+  'data-test-inline-error': true,
+  'data-test-inline-error-icon': true,
+
+  init() {
+    this._super(...arguments);
+    assert('[polaris-inline-error] Missing required `fieldID` param!', this.get('fieldID'));
```
andrewpye: The React component doesn't seem to care about this... why do we?

tests/integration/components/polaris-inline-error-test.js
```diff
@@ -0,0 +1,48 @@
...
+    });
+
+    await render(hbs`
+      {{polaris-inline-error
+        message=message
+        fieldID=fieldID
+      }}
+    `);
+
+    assert.dom('[data-test-inline-error]').exists();
```
andrewpye: These test selectors are pretty neat 😄

addon/components/polaris-inline-error.js
```diff
@@ -0,0 +1,35 @@
...
```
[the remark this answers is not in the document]
vladucu: was bitten by this the other day...it won't work in apps on pre-ember `3.1`
is there any way to support older versions of Ember too in the addon?

addon/templates/components/polaris-inline-error.hbs
```diff
@@ -0,0 +1,16 @@
+{{#if message}}
+  <div
+    class="Polaris-InlineError"
+    id={{concat fieldID "Error"}}
+    data-test-inline-error={{data-test-inline-error}}
```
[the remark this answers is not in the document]
vladucu: yes, but in a consuming app, in an acceptance test where you have multiple instances....would complicate testing no?

addon/components/polaris-inline-error.js
vladucu: the docs mention it's required...I'm fine with dropping this, tbh

andrewpye: True... don't see anything in the code enforcing it but maybe there's something in React that requires the `props` to be present 🤷‍♂️

vladucu: TypeScript

addon/templates/components/polaris-inline-error.hbs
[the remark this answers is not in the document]
sivakumar-kailasam: ah got it 🙂👍🏻. May be we could add a note in our readme doc somewhere under the testing section about this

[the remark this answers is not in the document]
vladucu: agreed...will add

PR #178, opened 2018-10-04
Ran ember-cli-update codemods

sivakumar-kailasam:
Nothing manual here, just ran all the codemods that come along with `ember-cli-update` by running `ember-cli-update --run-codemods`

- ember-modules-codemod
- ember-qunit-codemod
- ember-test-helpers-codemod
- es5-getter-ember-codemod


vladucu:
@sivakumar-kailasam I tried running this too....the codemods are not perfect and changes are still gonna be needed

was thinking we just run codemods when we go and update components too...this would also be easier to review...what do you think?

sivakumar-kailasam:
Sure sounds good!

Issue #177, opened 2018-10-04
Update polaris-breadcrumbs component


vladucu:
Changes [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0#diff-85d305f5099f6a7052e1b1bbdb5ea595R5)

andrewpye:
@vladucu maybe worth adding a link to the diff so whoever picks this up can see what needs to be updated?

vladucu:
good point...though it's already picked up 😉

vladucu:
fixed in #179

PR #179, opened 2018-10-04
Upgrade breadcrumbs component

tomnez:
Finishes https://github.com/smile-io/ember-polaris/issues/177

See the Polaris diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0?diff=split#diff-85d305f5099f6a7052e1b1bbdb5ea595)


addon/templates/components/polaris-breadcrumbs.hbs
```diff
@@ -1 +1,30 @@
-{{polaris-breadcrumbs/breadcrumb breadcrumb=breadcrumb}}
+{{#if breadcrumb.url}}
+  {{#polaris-unstyled-link
+    key=breadcrumb.content
```
andrewpye: IIRC this `key` property is a [React](https://reactjs.org/docs/lists-and-keys.html) thing so I don't think we need it

addon/templates/components/polaris-breadcrumbs.hbs
```diff
@@ -1 +1,30 @@
...
+    url=breadcrumb.url
+    class="Polaris-Breadcrumbs__Breadcrumb"
```
andrewpye: Attribute ordering?

addon/templates/components/polaris-breadcrumbs.hbs
```diff
@@ -1 +1,30 @@
...
+    <span data-test-breadcrumb-icon class="Polaris-Breadcrumbs__Icon">
+      {{polaris-icon source="chevron-left"}}
+    </span>
+    <span data-test-breadcrumb-content class="Polaris-Breadcrumbs__Content">
+      {{breadcrumb.content}}
+    </span>
+  {{/polaris-unstyled-link}}
+{{else}}
+  <button
+    key={{breadcrumb.content}}
```
andrewpye: Same here re. `key` and attribute ordering

addon/templates/components/polaris-breadcrumbs.hbs
```diff
@@ -1 +1,30 @@
...
+    type="button"
+  >
+    <span data-test-breadcrumb-icon class="Polaris-Breadcrumbs__Icon">
+      {{polaris-icon source="chevron-left"}}
+    </span>
+    <span data-test-breadcrumb-content class="Polaris-Breadcrumbs__Content">
+      {{breadcrumb.content}}
+    </span>
+  </button>
+{{/if}}
```
andrewpye: How do you feel about using a conditional to avoid duplicating the icon/content stuff between the link and button cases?

tests/integration/components/polaris-page-test.js
```diff
@@ -421,7 +420,7 @@ test('it handles breadcrumbs correctly', function(assert) {
 
   let breadcrumbLink = breadcrumbLinks[0];
   assert.ok(
-    breadcrumbLink.href.indexOf('/home/the-beginning/13/27') > -1,
+    breadcrumbLink.href.indexOf('/home/the-beginning') > -1,
```
andrewpye: IIRC this was previously testing that the breadcrumb handles models correctly - is this a capability we've lost now?

If that's the case, presumably host apps will need to refactor to use actions to transition between routes? In which case this is a breaking change so we'll need to be sure to document it and look at cutting a major release instead of a minor one...

andrewpye: ^^ follow-up thought to this, check if the documentation needs to change anywhere as a result of this. We don't have a docs page for `polaris-breadcrumbs` at the moment - it's included in the [`polaris-page` docs](https://github.com/smile-io/ember-polaris/blob/master/docs/page.md), which will probably need updating too.

andrewpye:
Don't forget to add a `CHANGELOG` entry for this 👍

tests/integration/components/polaris-page-test.js
andrewpye: Agreed that we should match the original component - this was one of the first things we added, and it was intended to be as ember-friendly in terms of links as possible 😉 But that's something we can deal with via actions now anyway.

I'm good with shipping a breaking change - better now than later when the user base has had a chance to grow...

addon/templates/components/polaris-breadcrumbs.hbs
andrewpye: Discussed this on slack with @tomnez and agreed that it's clearer to leave this as is

tomnez:
@andrewpye updated, had merge conflicts again so here's the docs change - https://github.com/smile-io/ember-polaris/pull/179/commits/15860362f6f098c3ec595e9d47a525ff0f788ad9

PR #181, opened 2018-10-05
Bump style guide & autofix files

sivakumar-kailasam:
We modified the style guide to have the ability to group single line & multi line fns (computed props) by their accessors(public vs private) along with other public & private props. This PR autofixes to follow it across files.

### DIFF
--- a/addon/components/polaris-drop-zone/file-upload.js
+++ b/addon/components/polaris-drop-zone/file-upload.js
@@ -20,13 +20,6 @@ export default Component.extend({
 
   layout,
 
-  iconDragDrop,
-  assetFileUpload,
-  assetImageUpload,
-
-  type: readOnly('context.type'),
-  size: readOnly('context.size'),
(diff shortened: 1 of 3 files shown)


addon/components/polaris-drop-zone/file-upload.js
```diff
@@ -53,6 +46,16 @@ export default Component.extend({
     return fileUpload[`actionHint${classify(type)}`];
   }),
 
+  iconDragDrop,
+
+  assetFileUpload,
+
+  assetImageUpload,
+
+  type: readOnly('context.type'),
```
vladucu: these seems to have been moved after multi-lines...any idea why?

sivakumar-kailasam: because those are public while these are private

sivakumar-kailasam:
Yep🙂

Issue #180, opened 2018-10-04
Upgrade polaris-button component


tomnez:
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0?diff=split#diff-80ce11d837170156710fe9e24d078a78)

vladucu:
fixed in #185

PR #185, opened 2018-10-05
Upgrade polaris-button component

tomnez:
Finishes https://github.com/smile-io/ember-polaris/issues/180

Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0?diff=split#diff-80ce11d837170156710fe9e24d078a78)

FYI I ran the `qunit-dom-codemod` on the test file here, so it'll be a lot easier to review the new test by just looking at this commit [here](https://github.com/smile-io/ember-polaris/pull/185/commits/9eadcbe4a7114f4be0ff4bbc1f9c757c279de09f?utf8=%E2%9C%93&diff=unified)

### DIFF
--- a/addon/components/polaris-button.js
+++ b/addon/components/polaris-button.js
@@ -156,7 +156,7 @@ export default Component.extend({
    *
    * @property icon
    * @public
-   * @type {SVG Source|Component}
+   * @type {String|Component}
    * @default null
    */
   icon: null,
--- a/addon/components/polaris-button.js
+++ b/addon/components/polaris-button.js
@@ -156,7 +156,7 @@ export default Component.extend({
    *
    * @property icon
    * @public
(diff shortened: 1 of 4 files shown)


addon/components/polaris-button.js
```diff
@@ -156,7 +156,7 @@ export default Component.extend({
    *
    * @property icon
    * @public
-   * @type {SVG}
+   * @type {SVG Source|Component}
```
andrewpye: This was my bad when I initially wrote this, but how do you feel about changing `SVG Source` to `String`, since that's the actual type of the `icon` property in `ember-polaris`?

tomnez: I also thought that, I figured it said SVG for a reason so I left it, but yes I will change 🕺

PR #189, opened 2018-10-08
Fix broken function required by polaris text field

sivakumar-kailasam:
Noticed this function while browsing through this component for a different effort.


sivakumar-kailasam:
If I could write that one true autofixer, then I'd be 

scrooge-in-money

Issue #188, opened 2018-10-05
Update polaris-choicelist component


tomnez:
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0?diff=split#diff-f9b830900da70d07e7b83b882b9e48d7)

tomnez:
https://github.com/smile-io/ember-polaris/pull/193

vladucu:
fixed in #193

PR #193, opened 2018-10-08
Upgrade v2.11.0 choicelist

tomnez:
Finishes https://github.com/smile-io/ember-polaris/issues/188

Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0?diff=split#diff-f9b830900da70d07e7b83b882b9e48d7)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,12 @@
 - [#168](https://github.com/smile-io/ember-polaris/pull/168) [INTERNAL] Upgrade to Ember CLI `3.4.2`
 - [#171](https://github.com/smile-io/ember-polaris/pull/171) [UPDATE] Shopify Polaris `v2.11.0`
 - [#174](https://github.com/smile-io/ember-polaris/pull/174) [FEATURE] Add `polaris-inline-error` component
-- [#179](https://github.com/smile-io/ember-polaris/pull/179) [FEATURE] Upgrade `polaris-breadcrumbs` to render button if `url` not present
-- [#187](https://github.com/smile-io/ember-polaris/pull/187) [FEATURE] Add `disabled` attribute to `polaris-choice`
-- [#185](https://github.com/smile-io/ember-polaris/pull/185) [FEATURE] Update `polaris-button` to accept components for `icon` value
+- [#179](https://github.com/smile-io/ember-polaris/pull/179) [ENHANCEMENT] Upgrade `polaris-breadcrumbs` to render button if `url` not present
+- [#187](https://github.com/smile-io/ember-polaris/pull/187) [ENHANCEMENT] Add `disabled` attribute to `polaris-choice`
+- [#185](https://github.com/smile-io/ember-polaris/pull/185) [ENHANCEMENT] Update `polaris-button` to accept components for `icon` value
 - [#191](https://github.com/smile-io/ember-polaris/pull/191) [ENHANCEMENT] Update `polaris-layout/annotated-section` styling to match Polaris v2.11.0
 - [#195](https://github.com/smile-io/ember-polaris/pull/195) [ENHANCEMENT] Add `nextKeys` and `previousKeys` attribute to `polaris-pagination`
+- [#193](https://github.com/smile-io/ember-polaris/pull/193) [ENHANCEMENT] Update `polaris-choice-list` to render `error` and `aria` attributes
 
 ### v.1.7.7 (october 8, 2018)
 - [#189](https://github.com/smile-io/ember-polaris/pull/189) [FIX] Detect length of decimal places in number as per expectations.
--- a/addon/components/polaris-choice-list.js
+++ b/addon/components/polaris-choice-list.js
@@ -112,6 +112,8 @@ export default Component.extend({
    */
   titleHidden: false,
 
+  'data-test-choice-list': true,
+
   /**
    * Callback when the selected choices change
    *
--- a/addon/components/polaris-choice-list.js
+++ b/addon/components/polaris-choice-list.js
@@ -4,6 +4,7 @@ import { bool } from '@ember/object/computed';
 import { guidFor } from '@ember/object/internals';
 import { isEmpty, isNone } from '@ember/utils';
 import ObjectProxy from '@ember/object/proxy';
+import { errorId } from '@smile-io/ember-polaris/utils/id';
 import layout from '../templates/components/polaris-choice-list';
 
 // Wrapper class to add an `isSelected` flag to the supplied choices.
@@ -24,8 +25,9 @@ export default Component.extend({
   tagName: 'fieldset',
 
   attributeBindings: [
-    'hasError:aria-invalid',
-    'finalNameError:aria-described-by',
+    'finalName:id',
+    'ariaInvalid:aria-invalid',
+    'ariaDescribedBy:aria-describedby',
   ],
 
   classNames: ['Polaris-ChoiceList'],
@@ -38,7 +40,7 @@ export default Component.extend({
    *
    * @property title
    * @public
-   * @type {string}
+   * @type {String}
    * @default null
    */
   title: null,
@@ -76,7 +78,7 @@ export default Component.extend({
    *
    * @property name
    * @public
-   * @type {string}
+   * @type {String}
    * @default null
    */
   name: null,
@@ -86,7 +88,7 @@ export default Component.extend({
    *
    * @property allowMultiple
    * @public
-   * @type {boolean}
+   * @type {Boolean}
    * @default false
    */
   allowMultiple: false,
@@ -96,8 +98,8 @@ export default Component.extend({
    *
    * @property error
    * @public
-   * @type {string|component}
-   * @default false
+   * @type {String|Component}
+   * @default null
    */
   error: null,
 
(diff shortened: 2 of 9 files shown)


addon/components/polaris-choice-list.js
```diff
@@ -104,6 +121,22 @@ export default Component.extend({
    */
   onChange() {},
 
+  /**
+   * @private
+   */
+  hasError: bool('error'),
```
andrewpye: Not sure it's a big deal, but the React implementation has this as `error != null` so maybe we should look at mimicking that behaviour here?

addon/components/polaris-choice-list.js
```diff
@@ -84,6 +91,16 @@ export default Component.extend({
    */
   allowMultiple: false,
 
+  /**
+   * Display an error message
+   *
+   * @property error
+   * @public
+   * @type {string|component}
```
andrewpye: These are from a time when I was young and foolish so I know this isn't your doing, but these types should be capitalised since they're class names

addon/components/polaris-choice-list.js
```diff
@@ -84,6 +91,16 @@ export default Component.extend({
...
+   * @default false
```
andrewpye: This `@default` doesn't match the actual default value 😉

addon/components/polaris-choice-list.js
```diff
@@ -104,6 +121,22 @@ export default Component.extend({
...
 
+  /**
+   * @private
+   */
+  hasError: bool('error'),
+
+  /**
+   * @private
+   */
+  finalNameError: computed('finalName', 'error', function() {
```
andrewpye: Wondering if this prop name is a bit unclear - what do you think about using something like `ariaDescribedBy`?

addon/components/polaris-choice-list.js
```diff
@@ -104,6 +121,22 @@ export default Component.extend({
...
+    if (isNone(this.get('error'))) {
+      return null;
+    }
+
+    return `${this.get('finalName')}Error`;
```
andrewpye: We have an existing [util](https://github.com/smile-io/ember-polaris/tree/master/addon/utils/id.js#L16) for generating this, maybe worth using that here? 🤷‍♂️

addon/components/polaris-choice-list.js
```diff
@@ -21,6 +22,12 @@ const CheckedChoice = ObjectProxy.extend({
  */
 export default Component.extend({
   tagName: 'fieldset',
+
+  attributeBindings: [
+    'hasError:aria-invalid',
+    'finalNameError:aria-described-by',
+  ],
```
andrewpye: We seem to be missing the `id` attribute that's been [added](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0?diff=split#diff-f9b830900da70d07e7b83b882b9e48d7R119) to the React version - the `fieldID` value we're passing to the new `polaris-inline-error` won't work without it

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
 - [#187](https://github.com/smile-io/ember-polaris/pull/187) [FEATURE] Add `disabled` attribute to `polaris-choice`
 - [#185](https://github.com/smile-io/ember-polaris/pull/185) [FEATURE] Update `polaris-button` to accept components for `icon` value
 - [#191](https://github.com/smile-io/ember-polaris/pull/191) [ENHANCEMENT] Update `polaris-layout/annotated-section` styling to match Polaris v2.11.0
-
+- [#193](https://github.com/smile-io/ember-polaris/pull/193) [FEATURE] Update `polaris-choice-list` to render `error` and `aria` attributes
```
vladucu: this is not a new feature 😜...update? (a few others here are marked as feature when they shouldn't, ex: button)

addon/components/polaris-choice-list.js
```diff
@@ -21,6 +22,12 @@ const CheckedChoice = ObjectProxy.extend({
  */
 export default Component.extend({
   tagName: 'fieldset',
+
+  attributeBindings: [
+    'hasError:aria-invalid',
+    'finalNameError:aria-described-by',
```
vladucu: should this be `aria-describedby`?

addon/components/polaris-choice-list.js
tomnez: Good call, I'm going to also update `hasError` to `ariaInvalid` since it's only used for the `aria-invalid` attribute binding 👍

tomnez:
@andrewpye @vladucu updated, had some conflicts so github might not be showing "since you last viewed" - the edits are all in [this commit](https://github.com/smile-io/ember-polaris/pull/193/commits/752389a41acf38f1bf67e5d94b1731d0df373a1c) if you want to check it on its own.

addon/components/polaris-choice-list.js
```diff
@@ -99,11 +117,29 @@ export default Component.extend({
...
+    return this.get('error') != null;
+  }),
+
+  /**
+   * @private
+   */
+  ariaDescribedBy: computed('finalName', 'error', function() {
+    if (isNone(this.get('error'))) {
+      return null;
+    }
```
andrewpye: This check doesn't seem to be in the [original](https://github.com/Shopify/polaris/blob/master/src/components/ChoiceList/ChoiceList.tsx#L121), any reason for adding it?

tomnez: Ahh no, I think this is a combination of my brain going rogue as well as a hilarious misunderstanding of what `aria-describedby` is for 🤦‍♂️

tomnez:
@vladucu @andrewpye updated with edits, as well as updated the test file to use built-in helpers and `ember-test-selectors` 👍

addon/templates/components/polaris-checkbox.hbs
```diff
@@ -9,6 +9,7 @@
 }}
   <span class="Polaris-Checkbox {{if error "Polaris-Checkbox--error"}}">
     <input
+      data-test-polaris-checkbox-input
```
vladucu: to be consistent with other places....let's rename this to omit `polaris`, just `data-test-checkbox-input`

addon/templates/components/polaris-choice-list.hbs
```diff
@@ -20,3 +20,12 @@
     </li>
   {{/each}}
 </ul>
+
+{{#if error}}
+  <div data-test-polaris-choice-list-error class="Polaris-ChoiceList__ChoiceError">
```
vladucu: same, remove `polaris`

addon/components/polaris-choice-list.js
```diff
@@ -99,11 +117,27 @@ export default Component.extend({
    *
    * @property onChange
    * @public
-   * @type {function}
+   * @type {Function}
    * @default noop
    */
   onChange() {},
 
+  'data-test-choice-list': true,
```
vladucu: just realized the other day too, having and then passing in something like `data-test-*` into a tagless component will result in a warning, so instead we should use a camelCased name
but, you're also not hooking this in with the first element in the template, so no value in having it if we're not using it...either hook it in or remove all together?

addon/components/polaris-choice/label.js
```diff
@@ -66,4 +66,6 @@ export default Component.extend({
    * @public
    */
   disabled: null,
+
+  'data-test-polaris-choice': true,
```
vladucu: same as the comment above

tests/integration/components/polaris-choice-list-test.js
```diff
@@ -1,586 +1,599 @@
...
-    'third choice - radio input has the correct name'
-  );
-  assert.notOk(radioInput.checked, 'third choice - radio input is not checked');
-});
+      }}
+    `);
+
+    const choices = findAll(choiceSelector);
+    const radioInputs = findAll(radioInputSelector);
+    assert.equal(choices.length, 3, 'renders three choices');
```
vladucu: you can just do `assert.dom(coiceSelector).exists({ count: 3 }, 'renders three choices');

tests/integration/components/polaris-choice-list-test.js
```diff
@@ -1,586 +1,599 @@
...
+
+    // Check the choices.
+    let choice = choices[0];
```
vladucu: no need to use these anymore, you can just use `assert.dom(`${choiceSelector}:nth-child(1))` to target first item and so on

tests/integration/components/polaris-choice-list-test.js
```diff
@@ -1,586 +1,599 @@
...
+    assert
+      .dom(choice)
+      .hasText('First option', 'first choice - has correct label text');
+    let radioInput = radioInputs[0];
```
vladucu: similar as above for the input

tests/integration/components/polaris-choice-list-test.js
tomnez: The `nth` selectors don't work in these cases because the choice selectors aren't all children of the same parent, so the test will only ever find `nth-child(1)` or `nth-of-type(1)` and then bork on trying to find the 2nd or 3rd, etc.

tomnez:
@vladucu made most of your edits, I updated selectors where I could but left a comment regarding the `nth` selectors.

Issue #194, opened 2018-10-08
Upgrade polaris-data-table component


tomnez:
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0?diff=split#diff-cb6bc8ab3e378297fe6e4026ca1118d1)

tomnez:
https://github.com/smile-io/ember-polaris/pull/197

vladucu:
fixed in #197

PR #197, opened 2018-10-09
Upgrade v2.11.0 data-table

tomnez:
Finishes https://github.com/smile-io/ember-polaris/issues/194

Pretty sizable makeover here, here's a screenshot from what the below code will render:

```hbs
{{#polaris-card}}
  {{polaris-data-table
    columnContentTypes=(array "text" "numeric" "numeric")
    headings=(array "Product" "Remaining stock" "Price")
    rows=(array
      (array "Kitten" 5 "$12,345")
      (array "Puppy" 3 "$6,789")
    )
    totals=(array "" 8 "$19,134")
    sortable=(array false false true)
    defaultSortDirection="descending"
    initialSortColumnIndex=1
    onSort=(action "sortData")
    footerContent="Showing 2 of 2 results"
  }}
{{/polaris-card}}
```

screen shot 2018-10-09 at 3 25 11 pm

### DIFF
--- a/addon/components/polaris-data-table.js
+++ b/addon/components/polaris-data-table.js
@@ -51,7 +51,9 @@ function measureColumn(tableData) {
 
 function isEdgeVisible(position, start, end) {
   let minVisiblePixels = 30;
-  return position >= start + minVisiblePixels && position <= end - minVisiblePixels;
+  return (
+    position >= start + minVisiblePixels && position <= end - minVisiblePixels
+  );
 }
 
 function getPrevAndCurrentColumns(tableData, columnData) {
@@ -331,19 +333,21 @@ export default Component.extend(
     },
 
     calculateColumnVisibilityData(collapsed) {
-      let {
-        table,
-        scrollContainer,
-        dataTable,
-      } = this.getProperties('table', 'scrollContainer', 'dataTable');
+      let { table, scrollContainer, dataTable } = this.getProperties(
+        'table',
+        'scrollContainer',
+        'dataTable'
+      );
 
       if (collapsed && table && scrollContainer && dataTable) {
         let headerCells = table.querySelectorAll('[class*=header]');
         let collapsedHeaderCells = Array.from(headerCells).slice(1);
         let fixedColumnWidth = headerCells[0].offsetWidth;
         let firstVisibleColumnIndex = collapsedHeaderCells.length - 1;
-        let tableLeftVisibleEdge = this.get('scrollContainer').scrollLeft + fixedColumnWidth;
-        let tableRightVisibleEdge = this.get('scrollContainer').scrollLeft + dataTable.offsetWidth;
+        let tableLeftVisibleEdge =
+          this.get('scrollContainer').scrollLeft + fixedColumnWidth;
+        let tableRightVisibleEdge =
+          this.get('scrollContainer').scrollLeft + dataTable.offsetWidth;
 
         let tableData = {
           fixedColumnWidth,
@@ -353,7 +357,7 @@ export default Component.extend(
         };
 
         let columnVisibilityData = collapsedHeaderCells.map(
-          measureColumn(tableData),
+          measureColumn(tableData)
         );
 
         let lastColumn = columnVisibilityData[columnVisibilityData.length - 1];
@@ -363,9 +367,10 @@ export default Component.extend(
             fixedColumnWidth,
             columnVisibilityData,
             isScrolledFarthestLeft: tableLeftVisibleEdge === fixedColumnWidth,
-            isScrolledFarthestRight: lastColumn.rightEdge <= tableRightVisibleEdge,
+            isScrolledFarthestRight:
+              lastColumn.rightEdge <= tableRightVisibleEdge,
           },
-          getPrevAndCurrentColumns(tableData, columnVisibilityData),
+          getPrevAndCurrentColumns(tableData, columnVisibilityData)
         );
       }
 
@@ -391,7 +396,7 @@ export default Component.extend(
         'footerContent',
         'truncate',
         'table',
-        'scrollContainer',
+        'scrollContainer'
       );
 
       let collapsed = false;
@@ -433,7 +438,7 @@ export default Component.extend(
         'footerContent',
         'truncate',
         'heights',
-        'table',
+        'table'
       );
       let rows = Array.from(table.getElementsByTagName('tr'));
 
@@ -489,7 +494,7 @@ export default Component.extend(
           'currentColumn',
           'previousColumn',
           'fixedColumnWidth',
-          'scrollContainer',
+          'scrollContainer'
         );
 
         if (!currentColumn || !previousColumn || !fixedColumnWidth) {
@@ -530,7 +535,7 @@ export default Component.extend(
           'initialSortColumnIndex',
           'sortDirection',
           'sortedColumnIndex',
-          'scrollContainer',
+          'scrollContainer'
         );
         sortedColumnIndex = isNone(sortedColumnIndex)
           ? initialSortColumnIndex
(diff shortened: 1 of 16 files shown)


addon/components/polaris-data-table/cell.js
```diff
@@ -46,14 +46,6 @@ export default Component.extend({
    */
   truncate: false,
 
-  /**
-   * @property presentational
-   * @type {boolean}
-   * @default false
-   * @public
-   */
-  presentational: false,
```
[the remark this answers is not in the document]
vladucu: also, I think we can drop this in `heading` template

addon/templates/components/polaris-data-table/heading.hbs
```diff
@@ -7,8 +7,8 @@
   truncate=truncate
   presentational=isPresentational
   defaultSortDirection=(if sortable defaultSortDirection)
-  sorted=(if sortable isSorted)
-  sortable=(if sortable isSortable)
+  sorted=(if sortable (eq sortedColumnIndex headingIndex))
+  sortable=(if sortable (get-array-item sortable headingIndex))
   sortDirection=(if sortable direction)
   onSort=(if sortable (action defaultOnSort index))
```
vladucu: should it be `headingIndex` instead of `index` here?

addon/components/polaris-data-table.js
```diff
@@ -405,12 +422,13 @@ export default Component.extend(
...
-      let { footerContent, truncate, heights } = this.getProperties(
+      let { footerContent, truncate, heights, table } = this.getProperties(
         'footerContent',
         'truncate',
-        'heights'
+        'heights',
+        'table'
       );
-      let rows = Array.from(this.get('table').getElementsByTagName('tr'));
+      let rows = Array.from(table.getElementsByTagName('tr'));
```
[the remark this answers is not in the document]
andrewpye: IIRC there was a difference in behaviour between the two, with `childNodes` throwing if I used it here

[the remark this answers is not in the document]
vladucu: thx @andrewpye 

@tomnez maybe we just need to check if there are childNodes before using it (`hasChildNodes()`) [source](https://developer.mozilla.org/en-US/docs/Web/API/Node/childNodes#Simple_usage)?

tomnez:
@vladucu updated!

addon/components/polaris-data-table.js
```diff
@@ -405,26 +429,30 @@ export default Component.extend(
...
+        if (!truncate) {
+          return (heights = rows.map((row) => {
+            let fixedCell = row.hasChildNodes() && row.childNodes;
+            return Math.max(row.clientHeight, fixedCell.clientHeight);
+          }));
+        }
 
-      return heights;
+        if (footerContent) {
+          let footerCellHeight = rows[rows.length - 1].children[0].clientHeight;
```
vladucu: let's be consistent...update here too to use `childNodes` and anywhere else we use this in these components

addon/templates/components/polaris-data-table/total.hbs
```diff
@@ -1,22 +1,12 @@
...
-    total=true
-    height=height
-    contentType="numeric"
-    text=total
-  }}
-{{/if}}
+{{polaris-data-table/cell
+  total=true
+  truncate=truncate
+  height=heights.firstObject
```
andrewpye: Should this be `heights[1]` instead of `heights[0]`?

(We have a helper that can allow you to do this as `(get-array-item heights 1)`)

addon/components/polaris-data-table.js
```diff
@@ -453,46 +481,40 @@ export default Component.extend(
...
+          scrollContainer.scrollLeft =
+            direction === 'right'
+              ? currentColumn.rightEdge - fixedColumnWidth
+              : previousColumn.leftEdge - fixedColumnWidth;
 
-          this.setProperties(
-            this.calculateColumnVisibilityData(this.get('collapsed'))
-          );
-        });
+          requestAnimationFrame(() => {
```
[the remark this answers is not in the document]
tomnez: I'll add it back 👍

addon/templates/components/polaris-data-table/total.hbs
tomnez: Eek good catch

tomnez:
@vladucu @andrewpye updated with the edits as well as swapped out selectors for `ember-test-selectors` in the test file.

addon/templates/components/polaris-data-table/cell.hbs
```diff
@@ -1,88 +1,52 @@
...
+{{#if (or header fixed)}}
   {{#if header}}
     <th
-      aria-hidden="true"
-      role="presentation"
-      id={{cellElementId}}
+      {{!-- template-lint-disable --}}
+      data-test-data-table-cell
+      data-test-data-table-cell-header
+      data-test-data-table-cell-sorted="{{sorted}}"
```
tomnez: This doesn't get assigned a boolean value in the test environment if I use 

```hbs
data-test-data-table-cell-sorted={{sorted}}
```

which is why it's in quotes with the linter disabled for this element. Same with the footer test-selector at the bottom of this files. I'm good with just checking the Polaris classes for these two and dropping the helpers if that sounds more reasonable.

andrewpye: I'm OK with this if it works 👍

tomnez: @vladucu we have tests that test that the correct number of cells were rendered, and we also have tests that test headers by themselves. That's why we have multiple selectors because not all cells are headers, etc.

tomnez:
@vladucu we don't have tests that really differentiate between row/data, the reason these all have a `data-test-data-table-cell` attribute is because the component is tagless and the template has a lot of conditionals so there's still only ever one `data-test-data-table-cell` that gets rendered at a time. I just looks a bit repetitive because of the template structure.

tests/integration/components/polaris-data-table-test.js
```diff
@@ -80,11 +82,14 @@ module('Integration | Component | polaris-data-table', function(hooks) {
...
-    assert.ok(
-      firstHeadingCell.classList.contains('Polaris-DataTable__Cell--sorted')
-    );
+    const firstHeadingCell = findAll(
+      '[data-test-data-table-cell][data-test-data-table-cell-header]'
+    )[0];
+
+    assert
+      .dom(firstHeadingCell)
+      .hasAttribute('data-test-data-table-cell-sorted', 'true');
```
vladucu: hmm, I don't think we should use selectors like this....we should use selectors just to match elements...in this case we just that the element has a test selector which is not what we want

here, for example, we should test that it has the sorted class....we could remove that class and the test would still pass, but functionality broken no?

vladucu: same in a bunch of other places around here

addon/templates/components/polaris-data-table/cell.hbs
vladucu: not sure I understand tbh or how this would help with this case (if I understand correctly lol)
if you want to easier select header cells/etc....just add a selector in the main data-table component so you can scope the selector to it....do you think something like that would work?

tests/integration/components/polaris-data-table-test.js
tomnez: Yeah I'm totally fine with testing that it has the sorted class, I just wasn't sure how hard we were enforcing the use of test-selectors so I was trying to use them for all cases.

addon/templates/components/polaris-data-table/cell.hbs
tomnez: Yeah that works too, I was just trying to put selectors on the elements we were actually selecting vs. using them as a scope and still using `th` or `td` in the selection path. I'll update this.

tomnez:
@vladucu updated, reverted back to using classes in the spots we discussed.

PR #199, opened 2018-10-10
Add support for disabled and loading states to polaris-setting-toggle

andrewpye:
Just found that the `disabled` and `loading` options in `polaris-setting-toggle`'s `action` hash weren't supported, so added them in.

### DIFF
--- a/docs/setting-toggle.md
+++ b/docs/setting-toggle.md
@@ -14,7 +14,6 @@ Inline usage:
   enabled=enabled
   action=(hash
     text="Toggle it!"
-    loading=isToggleSettingRunning
     onAction=(action "toggleSetting")
   )
 }}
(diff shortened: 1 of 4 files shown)


docs/setting-toggle.md
```diff
@@ -14,6 +14,7 @@ Inline usage:
   enabled=enabled
   action=(hash
     text="Toggle it!"
+    loading=isToggleSettingRunning
```
sivakumar-kailasam: should we document `disabled` too?

tests/integration/components/polaris-setting-toggle-test.js
```diff
@@ -160,3 +160,41 @@ test('it handles the suppied action correctly', function(assert) {
...
+        text="Toggle"
+        loading=isLoading
+        disabled=isDisabled
+        onAction=(action (mut dummy))
+      )
+    }}
+  `);
+
+  let button = find(settingActionButtonSelector);
+  assert.ok(
```
sivakumar-kailasam: this could also be written as 
```js
assert.dom(settingActionButtonSelector).hasClass('Polaris-Button--loading', 'button is in loading state when loading is true');
```

andrewpye: It could... but I want to get this out quickly so wasn't going to update the test file to use the newer things 🤷‍♂️

docs/setting-toggle.md
andrewpye: Thought about it but I'm not sure there's any real value

PR #201, opened 2018-10-10
Update polaris-form-layout component

### LINKED ISSUE #200: Update polaris-form-layout component
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0#diff-2c66129801b5f27907be29599dfc4cd9R2)

vladucu:
fix #200 

Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0#diff-2c66129801b5f27907be29599dfc4cd9R2)

### DIFF
--- a/addon/components/polaris-form-layout/group.js
+++ b/addon/components/polaris-form-layout/group.js
@@ -1,8 +1,15 @@
 import Component from '@ember/component';
 import layout from '../../templates/components/polaris-form-layout/group';
+import { isBlank } from '@ember/utils';
+import { computed } from '@ember/object';
+import { idVariation, helpTextId } from '../../utils/id';
 
 export default Component.extend({
-  attributeBindings: ['role'],
+  attributeBindings: [
+    'role',
+    'titleID:aria-labelledby',
+    'helpTextID:aria-describedby',
+  ],
 
   classNameBindings: ['condensed:Polaris-FormLayout--condensed'],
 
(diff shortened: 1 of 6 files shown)


addon/templates/components/polaris-form-layout/group.hbs
```diff
@@ -9,3 +13,7 @@
     {{text}}
   {{/if}}
 </div>
+
+{{#if helpText}}
+  <div id={{helpTextID}} class="Polaris-FormLayout__HelpText" data-test-form-layout-group-helpText>{{render-content helpText}}</div>
```
tomnez: nit: Do we want to break up the element blocks here for easier reading?

```hbs
<div id={{helpTextID}} class="Polaris-FormLayout__HelpText" data-test-form-layout-group-helpText>
  {{render-content helpText}}
</div>
```

If so, same with the title. But just a suggestion.

addon/utils/id.js
```diff
@@ -22,3 +22,8 @@ export function helpTextId(id) {
   return `${id}HelpText`;
 }
 export const computedHelpTextId = makeComputedIdGenerator(helpTextId);
+
+export function idVariation(id, suffix) {
+  return `${id}${suffix}`;
+}
+export const computedIdVariation = makeComputedIdGenerator(idVariation);
```
andrewpye: Now we've got this, how about refactoring the other helpers above to use `idVariation` instead so all of these utils go through the same function?

addon/components/polaris-form-layout/group.js
```diff
@@ -28,18 +35,58 @@ export default Component.extend({
...
    * @private
    */
   role: 'group',
 
+  'data-test-form-layout-group': true,
+
+  titleID: computed('title', function() {
+    if (isBlank(this.get('title'))) {
+      return;
+    }
```
andrewpye: Do you think it's worth wrapping up this logic in the `idVariation` util to clean these CPs up? I can't see that we'd ever want to generate an ID variation if the ID isn't present.

andrewpye: ooooh yeah, my bad! Plz ignore + continue as you were 😅

PR #211, opened 2018-10-12
Update text-style component

### LINKED ISSUE #210: Update text-style component
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0#diff-66a7abb8f39a28e40bea8a08b2609fb4R5)

vladucu:
fix #210
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0#diff-66a7abb8f39a28e40bea8a08b2609fb4R5)

### DIFF
--- a/addon/components/polaris-text-style.js
+++ b/addon/components/polaris-text-style.js
@@ -4,13 +4,36 @@ import { isEmpty } from '@ember/utils';
 import { classify } from '@ember/string';
 import layout from '../templates/components/polaris-text-style';
 
+const variationValue = {
+  positive: 'positive',
+  negative: 'negative',
+  strong: 'strong',
+  subdued: 'subdued',
+  code: 'code',
+};
+
+const variationElement = (variation) => {
+  switch (variation) {
+    case variationValue.code:
+      return 'code';
+    case variationValue.strong:
+      return 'b';
+    case variationValue.positive:
+    case variationValue.negative:
+    case variationValue.subdued:
+    default:
+      return 'span';
+  }
+};
+
 /**
  * Polaris text style component.
  * See https://polaris.shopify.com/components/titles-and-text/text-style
+ *
+ * @component polaris-text-style
  */
 export default Component.extend({
-  tagName: 'span',
-  classNameBindings: ['variationClass'],
+  tagName: '',
 
   layout,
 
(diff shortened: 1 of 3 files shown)


addon/templates/components/polaris-text-style.hbs
```diff
@@ -1,5 +1,26 @@
-{{#if hasBlock}}
-  {{yield}}
-{{else}}
-  {{text}}
+{{!-- NOTE: This is a bit hacky, but needed to support dynamically rendering different tag based on variation --}}
```
vladucu: not sure if there's a simple way to do this

addon/components/polaris-text-style.js
```diff
@@ -40,12 +63,24 @@ export default Component.extend({
    */
   text: null,
 
+  dataTestTextStyle: true,
```
vladucu: passing this for tagless component cause passing something like `data-test-*` will result in a warning

addon/templates/components/polaris-text-style.hbs
andrewpye: How about [`wrapper-element`](https://github.com/smile-io/ember-polaris/blob/master/addon/components/wrapper-element.js)? 🤔

vladucu: tried that, but it won't re-render if `elementTagName` changes

andrewpye: True... maybe use the approach in [`polaris-list`](https://github.com/smile-io/ember-polaris/blob/master/addon/templates/components/polaris-list.hbs#L2) then?

andrewpye: Still not perfect, but IMO an improvement 🤷‍♂️

andrewpye: True... maybe use the approach in [`polaris-list`](https://github.com/smile-io/ember-polaris/blob/master/addon/templates/components/polaris-list.hbs#L2) then? Still not perfect but maybe an improvement 🤷‍♂️

vladucu: yeah, guess we can use that

addon/components/polaris-text-style.js
```diff
@@ -4,13 +4,36 @@ import { isEmpty } from '@ember/utils';
 import { classify } from '@ember/string';
 import layout from '../templates/components/polaris-text-style';
 
+const variationValue = {
```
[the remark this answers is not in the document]
vladucu: true that this is Polaris uses in 2.11, though in 2.12 they use this...so just _traveled in the future_ 😄

andrewpye:
Actually just noticed the failing tests, fix that and we're good 😉

PR #216, opened 2018-10-22
Update labelled to v2.11.0

### LINKED ISSUE #215: Update labelled component to v2.11.0
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0#diff-9051569aebf31a5aac93f522a9587df3L8)

andrewpye:
Closes #215

I didn't update to use `ember-test-selectors` here simply because the existing tests are just straight ports of the React implementation's tests. Let me know if you guys feel like we should make that update anyway and I'll do it :+1:

### DIFF
--- a/addon/components/polaris-labelled.js
+++ b/addon/components/polaris-labelled.js
@@ -40,7 +40,7 @@ export default Component.extend({
   /**
    * Error to display beneath the label
    *
-   * @type {String|Component}
+   * @type {String|Component|Boolean}
    * @public
    */
   error: null,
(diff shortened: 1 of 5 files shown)


addon/components/polaris-labelled.js
```diff
@@ -87,4 +87,15 @@ export default Component.extend({
...
   helpTextId: computedHelpTextId('id').readOnly(),
+
+  /**
+   * Flag indicating whether to render the error component
+   *
+   * @type {Boolean}
+   * @private
+   */
+  shouldRenderError: computed('error', function() {
+    let { error } = this;
```
vladucu: I think we still want to use Ember getters to support older versions of Ember.....from what I remember I run into issues trying to do this before

addon/templates/components/polaris-labelled.hbs
```diff
@@ -20,12 +20,9 @@
 
   {{yield}}
 
-  {{#if error}}
+  {{#if shouldRenderError}}
     <div id={{errorId}} class="Polaris-Labelled__Error">
```
vladucu: `id={{errorId}}` seems no longer needed here

andrewpye: Good catch, thanks!

PR #218, opened 2018-10-22
Update page to v2.11.0

### LINKED ISSUE #217: Update page component to v2.11.0
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0#diff-3887dd68aca2db017df62147a97a89e9R1)

andrewpye:
Closes #217 by adding `titleMetadata` and `primary` support to page header.

Screen grab with `titleMetadata=(component "polaris-badge" status="success" text="Metadata")` and `primaryAction=(hash primary=false ...)`:

image

### DIFF
--- a/addon/components/polaris-page/header.js
+++ b/addon/components/polaris-page/header.js
@@ -25,6 +25,16 @@ export default Component.extend({
   title: null,
 
   /**
+   * Important and non-interactive status information shown immediately after the title
+   *
+   * @property titleMetadata
+   * @public
+   * @type {String|Component}
+   * @default null
+   */
+  titleMetadata: null,
+
+  /**
    * Visually hide the title
    *
    * @property titleHidden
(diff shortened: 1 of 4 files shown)


addon/components/polaris-page/header.js
```diff
@@ -104,4 +115,11 @@ export default Component.extend({
   hasActions: or('primaryAction', 'secondaryActions').readOnly(),
   hasSecondaryActions: gt('secondaryActions.length', 0).readOnly(),
   hasRollup: gt('secondaryActions.length', 1).readOnly(),
+
+  shouldRenderPrimaryActionAsPrimary: computed('primaryAction.primary', function() {
```
andrewpye: Very happy to rename this if anyone can think of anything better :wink:

tomnez: Yeah, it's weird that this is a thing now but 🤷‍♂️ I'm good with it

vladucu:
Would be nice to drop in a quick test, nothing fancy, just that we render these properties correctly
Everything else looks good though

andrewpye:
@sivakumar-kailasam @vladucu added a couple of tests, not run codemods on them though because it causes issues with registering the mock `svg-jar` helper and I figured we could use our time more productively at this point

Issue #198, opened 2018-10-09
Update polaris-date-picker and polaris-date-picker/month components


tomnez:
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0?diff=split#diff-31dd30c75130536eea45b3bccfbb92b9)

vladucu:
fixed in #219

PR #219, opened 2018-10-22
Upgrade v2.11.0 date picker

tomnez:
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0?diff=split#diff-31dd30c75130536eea45b3bccfbb92b9)

Finishes https://github.com/smile-io/ember-polaris/issues/198

### DIFF
--- a/addon/components/polaris-date-picker.js
+++ b/addon/components/polaris-date-picker.js
@@ -11,13 +11,11 @@ import {
   getNextDisplayMonth,
   getPreviousDisplayYear,
   getPreviousDisplayMonth,
-  weekdays, // obj
+  weekdays,
   isSameDay,
 } from '../utils/dates';
 
-// TODO: add changes for Polaris v2.2.0
-// See https://github.com/Shopify/polaris/compare/v2.0.0...v2.2.0?diff=split#diff-31dd30c75130536eea45b3bccfbb92b9
-// (fixes https://github.com/Shopify/polaris/issues/313, if the bug's present in our Ember implementation)
+const weekStartsOn = weekdays.Sunday;
 
 export default Component.extend({
   classNames: ['Polaris-DatePicker'],
@@ -83,6 +81,16 @@ export default Component.extend({
    * @default false
    */
   multiMonth: false,
+
+  /**
+   * First day of week. Sunday by default
+   *
+   * @property weekStartsOn
+   * @public
+   * @type {String}
+   * @default 'Sunday'
+   */
+  weekStartsOn,
 
   /**
    * Callback when date is selected
@@ -353,14 +361,4 @@ export default Component.extend({
       this.set('focusDate', null);
     },
   },
-
-  /**
-   * First day of week. Sunday by default
-   *
-   * @property weekStartsOn
-   * @public
-   * @type {String}
-   * @default 'Sunday'
-   */
-  weekStartsOn: weekdays.Sunday,
 });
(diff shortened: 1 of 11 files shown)


addon/components/polaris-date-picker.js
```diff
@@ -10,6 +11,8 @@ import {
   getNextDisplayMonth,
   getPreviousDisplayYear,
   getPreviousDisplayMonth,
+  weekdays, // obj
```
andrewpye: Is the `// obj` comment needed here?

addon/components/polaris-date-picker.js
```diff
@@ -10,6 +11,8 @@ import {
...
+  isSameDay,
 } from '../utils/dates';
 
 // TODO: add changes for Polaris v2.2.0
```
andrewpye: How 'bout these guys? Are those changes present now?

addon/components/polaris-date-picker.js
```diff
@@ -316,4 +353,14 @@ export default Component.extend({
...
+
+  /**
+   * First day of week. Sunday by default
+   *
+   * @property weekStartsOn
+   * @public
+   * @type {String}
+   * @default 'Sunday'
+   */
+  weekStartsOn: weekdays.Sunday,
```
andrewpye: This guy should probs be up with the rest of the public interface 😉

addon/components/polaris-date-picker/weekday.js
```diff
@@ -33,4 +33,5 @@ export default Component.extend({
    * @default false
    */
   current: false,
+  'data-test-date-picker-weekday': true,
```
andrewpye: Nit: blank line before this?

addon/utils/dates.js
```diff
@@ -65,15 +65,20 @@ export function abbreviationForWeekday(weekday) {
...
   let currentWeek = weeks[0];
   let currentDate = firstOfMonth;
 
-  for (let i = 0; i < firstDayOfWeek; i++) {
+  const orderedWeekday = getWeekdaysOrdered(weekStartsOn);
+  for (
+    let i = 0;
+    i < orderedWeekday.indexOf(weekdaysArray[firstDayOfWeek]);
+    i++
+  ) {
```
andrewpye: OMG thx prettier 😏

addon/components/polaris-date-picker.js
tomnez: Ah! That was supposed to be a temporary reminder to myself since Polaris uses enums for weekdays haha. Will remove..

addon/components/polaris-date-picker.js
tomnez: @andrewpye Ah meant to comment on this, the linter hates these types of properties (it considers this an "unknown property type") and wants everything above it. I tried ignoring with eslint and prettier comments but it didn't seem to work. Any ideas on working around this?

addon/utils/dates.js
tomnez: 😞

addon/components/polaris-date-picker.js
andrewpye: Unfortunately I don't, but maybe @sivakumar-kailasam does?

addon/components/polaris-date-picker.js
tomnez: Ah yeah this can be removed. Ironically the additions from 2.0->2.2 are stripped out in 2.11 (other than translations on accessibility labels which we don't have)

addon/components/polaris-date-picker.js
tomnez: Looks like assigning to a variable first fixes it. Kinda wack, but better than having it at the bottom all willy nilly (I think)

Issue #223, opened 2018-10-23
Upgrade drop-zone component to v2.11.0


tomnez:
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0#diff-7ee058419e736ffe4c643519496b920f)

PR #225, opened 2018-10-23
Upgrade dropzone to v2.11.0

tomnez:
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0#diff-7ee058419e736ffe4c643519496b920f)

Finishes https://github.com/smile-io/ember-polaris/issues/223


addon/components/empty-wrapper.js
```diff
@@ -0,0 +1,8 @@
+import Component from '@ember/component';
+import layout from '../templates/components/empty-wrapper';
+
+export default Component.extend({
+  tagName: '',
+
+  layout,
```
tomnez: Curious as to what you all think of this, it's a workaround for creating templates like this:

```tsx
const labelledDropzoneMarkup = label ? (
  <Labelled
    id={id}
    label={label}
    action={labelAction}
    labelHidden={labelHidden}
  >
    {dropZoneMarkup}
  </Labelled>
) : (
  dropZoneMarkup
);
```

So we could use the `with` helper in templates like so:

```hbs
{{#with (if label
  (component "polaris-labelled"
    id=id
    label=label
    labelHidden=labelHidden
    action=labelAction
  )
  (component "empty-wrapper")
) as |wrapper|}}
  {{#component wrapper}}
    {{!-- dropzone markup here --}}
  {{/component}}
{{/with}}
```

without having to duplicate template logic, etc.

andrewpye: We've built a `no-op` component in another addon which basically serves the same purpose, so perhaps we could look at moving that here?

Alternatively (and possibly better), I don't see any reason why using our existing `wrapper-element` component with `tagName=""` wouldn't work, could you give it a try?

addon/components/polaris-drop-zone.js
```diff
@@ -4,6 +4,7 @@ import { or } from '@ember/object/computed';
 import { classify } from '@ember/string';
 import { throttle, scheduleOnce } from '@ember/runloop';
 import { isNone, isPresent } from '@ember/utils';
+import { createUniqueIDFactory } from '../utils/create-unique-id-factory';
```
andrewpye: We've eschewed this in the past in favour of using Ember's unique IDs via `guidFor(this)`, is there any reason not to continue that path here?

addon/components/polaris-drop-zone.js
```diff
@@ -19,18 +20,47 @@ const iconDragDrop = 'drag-drop';
...
-  classNames: ['Polaris-DropZone'],
+  /**
+   * ID for file input
+   *
+   * @type {String}
+   * @default null
+   * @public
+   * @property id
+   */
+  id: null,
```
andrewpye: Not 100% sure how Ember handles passing `id=<whatever>` into a component, but from a faint memory I think it might use that as the `id` of that component's element, which will cause problems if we then try using that same ID on another element. With that in mind, we might be safer renaming this to something like `inputId` (as I think we've done for a bunch of other form-related components) and documenting the difference from the React interface in the component's markdown file.

You could also default this to a CP that returns `guidFor(this)`, which I think we've also done elsewhere, if that cleans up logic later on.

addon/components/polaris-drop-zone.js
```diff
@@ -477,31 +539,40 @@ export default Component.extend(ContextBoundEventListenersMixin, {
       return;
     }
 
+    let dropzoneContainer = this.element.querySelector('.Polaris-DropZone');
```
andrewpye: Since we're using this in a couple of places, how do you feel about making a `dropzoneContainer` CP?

addon/templates/components/polaris-drop-zone.hbs
```diff
@@ -1,60 +1,79 @@
...
+    action=labelAction
+  )
+  (component "empty-wrapper")
+) as |wrapper|}}
+  {{#component wrapper}}
+    <div
+      data-test-drop-zone
+      class={{dropZoneClasses}}
+      aria-disabled={{ariaDisabled}}
+    >
```
andrewpye: Just curious as this is a noticeable difference from the React source, is there any reason for not applying the click/drag start actions directly on this element rather than attaching them in the JS file?

addon/utils/create-unique-id-factory.js
```diff
@@ -0,0 +1,4 @@
+export function createUniqueIDFactory(prefix) {
+  let index = 1;
+  return () => `${prefix}${index++}`;
+}
```
andrewpye: Can possibly remove this file because of my other comment about using `guidFor`

addon/components/polaris-caption.js
```diff
@@ -20,4 +20,6 @@ export default Component.extend({
    * @default: null
    */
   text: null,
+
+  'data-test-caption': true,
```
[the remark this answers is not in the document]
vladucu: agree this is not necessary, but nice to have, ex: a component doesn't have this and another one that uses it is being tested...you can't really add to the instance

addon/components/polaris-drop-zone.js
vladucu: I think we're setting this as `this.node` (in `setupNode()`) before this runs, so we should just use that everywhere else?

addon/templates/components/polaris-drop-zone.hbs
```diff
@@ -1,60 +1,79 @@
...
-    accept={{accept}}
-    disabled={{disabled}}
-    multiple={{allowMultiple}}
-    onchange={{action handleDrop}}
-  >
-{{/polaris-visually-hidden}}
+      {{#polaris-visually-hidden}}
+        <input
+          data-test-drop-zone-hidden-input
+          id={{concat elementId "-input"}}
```
vladucu: I think we've missed updating here to use the `id`

addon/components/polaris-drop-zone.js
vladucu: I think it should be fine setting the `id` as long as it's unique (we don't re-use it in this component on multiple components - which I don't think we do)
setting a default seems to be done in `getDerivedStateFromProps()`, but agree we have `guidFor` instead of `createUniqueIDFactory`

addon/components/polaris-drop-zone.js
```diff
@@ -559,5 +636,15 @@ export default Component.extend(ContextBoundEventListenersMixin, {
...
     this.set('state.error', this.get('error'));
     this.setupEvents();
+
+    if (this.get('openFileDialog')) {
+      this.triggerFileDialog();
+    }
+  },
+
+  didUpdateAttrs() {
+    if (this.get('openFileDialog')) {
```
vladucu: let's call `_super` here....you never know what we do in the future

addon/templates/components/polaris-drop-zone.hbs
vladucu: also, in JS, we'll want to tweak `fileInputNode` CP?

tests/integration/components/polaris-drop-zone-test.js
```diff
@@ -1237,5 +1229,39 @@ module('Integration | Component | polaris-drop-zone', function(hooks) {
...
+          this.set('actionFired', true);
+        },
+      });
+
+      await render(hbs`
+        {{polaris-drop-zone
+          label="my label"
+          labelAction=(hash
+            text="my label action"
+            onAction=labelAction
```
vladucu: `onAction=(action labelAction)`

tests/integration/components/polaris-drop-zone-test.js
```diff
@@ -1237,5 +1229,39 @@ module('Integration | Component | polaris-drop-zone', function(hooks) {
...
+          )
+        }}
+      `);
+
+      assert
+        .dom(dropZoneLabelWrapperSelector)
+        .exists('A labelled component is wrapped around the dropzone');
+      assert.notOk(
```
vladucu: no harm here, though this is unnecessary....this is basically testing `polaris-labelled` component not firing the action on render....likely just that it fires on click should be enough

addon/components/polaris-caption.js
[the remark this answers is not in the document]
andrewpye: Not sure I follow your example @vladucu but I don't feel particularly strongly about this either way

addon/components/empty-wrapper.js
tomnez: Ah I saw `wrapper-element` but didn't think to try passing a blank tag name. I'll see if that works 👍

addon/templates/components/polaris-drop-zone.hbs
tomnez: Ah you're right - I originally left it in the component file because that's where it lived originally, but that was because this element used to be the element generated by the component. Now it's available in the template so yes I can move it there 👍

addon/components/polaris-drop-zone.js
tomnez: Ooh embarrassing 😳

tests/integration/components/polaris-drop-zone-test.js
tomnez: Yeah I was just being super cautious about getting a false positive or anything. I'll remove it, this whole action test is likely not necessary since it should be covered in the labelled tests, but I'll at least keep the second assertion.

vladucu: sounds good to me!

tomnez:
Updated. Notes from the updates:

- Left test-selectors as-is
- Left `id` as `id`, but used `guidFor` instead of the ID util

addon/components/polaris-drop-zone.js
andrewpye: Yeah I know, just that so far we've avoided doing that and I'd like to keep our addon self-consistent. We can always look at switching all of the `guidFor`s across to follow the React implementation at a later stage :+1

vladucu: yeah, I think this is not really a problem, but more an advantage that we use Ember 😉

PR #226, opened 2018-10-24
Add sticky component

### LINKED ISSUE #224: Add Sticky component
We need this to be able to port the `ResourceList` component across from React-land 😄 

Original `Sticky` component source is [here](https://github.com/Shopify/polaris/tree/master/src/components/Sticky). We will also need a `StickyManager` service, React source [here](https://github.com/Shopify/polaris/blob/master/src/components/AppProvider/StickyManager.ts).

andrewpye:
Closes #224 by adding `polaris-sticky` and the associated `sticky-manager` service. Original source is [here](https://github.com/Shopify/polaris/tree/master/src/components/Sticky) for the component and [here](https://github.com/Shopify/polaris/blob/master/src/components/AppProvider/StickyManager.ts) for the service.

I've added a test page to the dummy app - here's a GIF of it in action:
polaris-sticky

You can compare it to the same page in React world [here](https://pmonqjp3q7.codesandbox.io/) - they seem to behave identically for all intents and purposes 🙌


addon/components/polaris-color-picker/slidable.js
```diff
@@ -4,6 +4,7 @@ import { computed } from '@ember/object';
 import { isNone, typeOf } from '@ember/utils';
 import { htmlSafe } from '@ember/string';
 import layout from '../../templates/components/polaris-color-picker/slidable';
+import { getRectForNode } from '@smile-io/ember-polaris/utils/geometry';
```
andrewpye: I'd previously ignored the React `getRectForNode` util because in our usage up to this point it's only ever been calling `getBoundingClientRect` on the passed-in node, but now that `sticky-manager` passes the `document` element (which doesn't support `getBoundingClientRect`) I figured we should match what the React implementation does everywhere, to make future updates cleaner.

addon/components/polaris-sticky.js
```diff
@@ -0,0 +1,160 @@
...
+      'placeHolderNode',
+      'offset',
+      'boundingElement',
+      'disableWhenStacked'
+    );
+    stickyManager.registerStickyItem(
+      {
+        stickyNode,
+        placeHolderNode,
+        handlePositioning: (...positioningArgs) => this.handlePositioning(...positioningArgs),
```
andrewpye: This is slightly different to the React implementation, because they use an `autobind` decorator on `handlePositioning` which takes care of context binding for them.

addon/utils/geometry.js
```diff
@@ -0,0 +1,17 @@
+export function getRectForNode(node) {
+  if (!(node instanceof Element)) {
+    return {
+      top: 0,
+      left: 0,
```
andrewpye: These `top` and `left` values default to `0` in the React version of this, but we need to specify them explicitly here otherwise they're undefined which breaks the sticky positioning.

addon/services/sticky-manager.js
```diff
@@ -0,0 +1,245 @@
...
+        ({ stickyNode }) => node === stickyNode
+      );
+
+      return nodeFound >= 0;
+    },
+
+    init() {
+      this._super(...arguments);
+
+      let container = this.get('container') || document;
```
andrewpye: The original `StickyManager` React code doesn't default this to `document`, but the React `AppProvider` _does_ [set it to `document`](https://github.com/Shopify/polaris/blob/master/src/components/AppProvider/AppProvider.tsx#L48) if `document` exists.

addon/services/sticky-manager.js
```diff
@@ -0,0 +1,245 @@
...
+  return node === document;
+}
+
+function scrollTopFor(container) {
+  return isDocument(container)
+    ? document.body.scrollTop || document.documentElement.scrollTop
+    : container.scrollTop;
+}
+
+function horizontallyOverlaps(rect1, rect2) {
```
sivakumar-kailasam: assuming you copied this over from the react implementation? How do you feel about renaming them with string prefixes as `firstRect` and `secondRect`? probably reduces chance for typos

addon/components/polaris-popover/content.js
```diff
@@ -82,7 +82,7 @@ export default Component.extend({
       return;
     }
     const paneContent = pane.firstElementChild;
-    const paneContentRect = paneContent.getBoundingClientRect();
+    const paneContentRect = getRectForNode(paneContent);
     Ember$('div.Polaris-Popover__Content', content).css({
```
sivakumar-kailasam: may use the native api instead of jQuery?

andrewpye: I think that's beyond the scope of this PR since this was just a quick swap of an old util for a new one... we should look into removing JQuery but have hit difficulties doing so before.

addon/services/sticky-manager.js
andrewpye: I'd like to keep the naming the same so future diffs with the React code are clearer

addon/components/polaris-popover/content.js
sivakumar-kailasam: Sounds good

addon/services/sticky-manager.js
```diff
@@ -0,0 +1,245 @@
...
+    getOffset(node) {
+      let stuckItems = this.get('stuckItems');
+      let stuckItemsLength = stuckItems.get('length');
+      if (stuckItemsLength === 0) {
+        return 0;
+      }
+
+      let offset = 0;
+      let count = 0;
+      let stuckNodesLength = stuckItemsLength;
```
vladucu: any reason for not just declaring `stuckItemsLength` as `stuckNodesLength` and drop this?

addon/services/sticky-manager.js
vladucu: can you add this comment in code pls?

addon/services/sticky-manager.js
```diff
@@ -0,0 +1,245 @@
...
+      let left = getRectForNode(placeHolderNode).left;
+
+      let sticky;
+
+      if (boundingElement == null) {
+        sticky = scrollPosition >= placeHolderNodeCurrentTop;
+      } else {
+        let stickyItemHeight = getRectForNode(stickyNode).height;
+        let stickyItemBottomPosition =
+          getRectForNode(boundingElement).bottom -
```
vladucu: should we still be using here `getBoundingClientRect` instead of `getRectForNode `....like React does? I don't think `getRectForNode` has `bottom` property, no?

addon/services/sticky-manager.js
andrewpye: Ah this is just my bad 😛

addon/utils/geometry.js
andrewpye: TypeScript I think...

addon/utils/geometry.js
```diff
@@ -0,0 +1,17 @@
...
+      width: window.innerWidth,
+      height: window.innerHeight,
+    };
+  }
+  var rect = node.getBoundingClientRect();
```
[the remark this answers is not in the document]
andrewpye: That's what's in the React code ¯\\\_(ツ)_/¯

addon/utils/geometry.js
vladucu: I realise we didn't do this before, but I don't the fact that the code uses TypeScript should be a problem....since that code will be compiled to JS no?
likely adding `@shopify/javascript-utilities` as a dep and  `import { getRectForNode } from '@shopify/javascript-utilities/geometry';` should work out of the box

addon/utils/geometry.js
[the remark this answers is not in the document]
andrewpye: yeah I was looking at the transpiled stuff 🤦‍♂️

addon/utils/geometry.js
andrewpye: Seems to be working, just pushed the changes 👍

vladucu:
nice!!
🚀

andrewpye:
Merging this into `dev` despite the test error on Ember 2.12, since we'll drop 2.12 support before merging `dev` down.

PR #227, opened 2018-10-24
Update text-field component

### LINKED ISSUE #212: Update text-field component
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0#diff-ce1bd24f5c7c2bd4b9fb5d55a5aa97aeR6)

vladucu:
fixes #212
Diff [here](https://github.com/Shopify/polaris/compare/v2.2.0...v2.11.0#diff-ce1bd24f5c7c2bd4b9fb5d55a5aa97aeR6)

### DIFF
--- a/addon/components/polaris-text-field.js
+++ b/addon/components/polaris-text-field.js
@@ -646,8 +646,6 @@ export default Component.extend(
         let { key } = event;
         let type = this.get('type');
         let numbersSpec = /[\d.eE+-]$/;
-        // eslint-disable-next-line
-debugger;
 
         if (
           type !== 'number' ||
--- a/addon/components/polaris-text-field.js
+++ b/addon/components/polaris-text-field.js
@@ -9,6 +9,7 @@ import { isPresent } from '@ember/utils';
 import ContextBoundTasksMixin from 'ember-lifeline/mixins/run';
 import ContextBoundEventListenersMixin from 'ember-lifeline/mixins/dom';
 import layout from '../templates/components/polaris-text-field';
+import { normalizeAutoCompleteProperty } from '../utils/normalize-auto-complete';
 
 const allowedTypes = [
   'text',
@@ -39,7 +40,7 @@ export default Component.extend(
   ContextBoundTasksMixin,
   ContextBoundEventListenersMixin,
   {
-    classNameBindings: ['labelHidden:Polaris-Labelled--hidden'],
+    tagName: '',
 
     layout,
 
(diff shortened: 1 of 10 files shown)


andrewpye:
@vladucu there's a test failure that needs fixing for this

vladucu:
yeah, not done yet...will assign you once done

andrewpye:
Oooh my bad, I got the email about the PR and assumed it was a review request 🤦‍♂️

vladucu:
no worries 😉

addon/components/polaris-text-field.js
```diff
@@ -403,6 +463,15 @@ export default Component.extend(
       }
     ).readOnly(),
 
+    inputClassName: computed('suffix', function() {
+      let classes = ['Polaris-TextField__Input'];
+      if (this.get('suffix')) {
```
sivakumar-kailasam: Why not use an if helper in the template?

vladucu: if there's a reason, I guess it's because React implementation has it like this....but yeah, we can totally do that, just wondering if more things will be added here in the future or not

tomnez: This follows suit with how we apply classes to all our other components, I think it would be fine to keep this as-is.

addon/components/polaris-text-field.js
```diff
@@ -541,8 +624,9 @@ export default Component.extend(
         this.set('height', height);
       },
 
-      handleChange(e) {
-        this.get('onChange')(e.target.value, this.get('id'));
+      handleChange(event) {
+        // event.preventDefault();
```
tomnez: Re-enable if needed, otherwise delete 😛

addon/components/polaris-text-field.js
```diff
@@ -348,7 +400,15 @@ export default Component.extend(
      */
     onBlur() {},
 
+    /**
+     * @private
+     */
+    focus: false,
+
+    dataTestTextField: 'text-field',
```
tomnez: Any reason some of these files are formatted this way and other are using snakecase with `true` values? If not we should decide on sticking to one format.

vladucu: yes
I think I mentioned this before....the reason is that for tag-less components, passing `data-test-*` format will cause ember-test-selectors to throw a warning (since you can't add something if we don't have an element for the component)....so instead we pass camelCased format and bind it to the first element as a test seletor in the tag-less component template

let me know if that doesn't makes sense

tomnez: Ahh yes I forgot. My bad 😞

vladucu: haha, no harm in asking 😉

Issue #162, opened 2018-08-01
Action list item's button submits parent form


tomnez:
In the following scenario, the button click on the `polaris-action-list/item` will submit the form that the list is nested inside of:

```hbs
<form onsubmit={{action "submitted"}}>
  {{polaris-action-list
    items=(array
      (hash
        text="Click me"
        onAction=(action "onClick")
      )
    )
  }}
</form>
```

```javascript
import Controller from '@ember/controller';

export default Controller.extend({
  actions: {
    onClick() {
      alert('action list item clicked.')
    },

    submitted() {
      alert('form submitted.');
    }
  }
});
```

The page also reloads after the form submits. The above implementation is not likely to be used, but this has also happened in other popular form addons such as `ember-form-for`.

As a quick test, I added an `event.preventDefault();` [here](https://github.com/smile-io/ember-polaris/blob/master/addon/components/polaris-action-list/section.js#L32) in the `onItemAction` and it seemed to fix the issue.

Demo gif:
![form](https://user-images.githubusercontent.com/2292367/43547447-3daf4e8c-95a9-11e8-8ed6-7474b06fb885.gif)

andrewpye:
@vladucu @tomnez yep, managed to repro this on the React component as well, I'll submit a bug report 👍

andrewpye:
@tomnez this was fixed in the React implementation in v2.6.0, can you confirm whether our updates have fixed the problem for us too?

tomnez:
@andrewpye looks like our latest `#dev` branch still has this issue. I don't see any specific code in the React implementation that prevents the event from bubbling up, should we add an `event.preventDefault();` as proposed?

We already have an `event.stopPropagation();` which is also not in the React implementation so it seems like we're already adding some extra logic to match the React behavior, I don't think it would hurt to add another line to get rid of this bug.

Also FWIW it looks like our implementation of these 3 components (ActionList, Section, and Item) is pretty different from the React version. I wonder if it's worth overhauling to see if it fixes the problem.

cc @vladucu

vladucu:
@tomnez taking everything into account, I think best approach right now would be:

- quick fix now - prevent the default
- create an issue to refactor action list

PR #237, opened 2018-10-25
Fix action item actions event bubbling

tomnez:
Fix for https://github.com/smile-io/ember-polaris/issues/162


tomnez:
Added this to `dev` since we're closing in on the v2.11.0 branch, I figure it can be a part of that release rather than go through a release cycle for this fix and then another one a couple days later after v2.11.0.

Let me know if anyone thinks I should repurpose this to go straight to `master` for a quick bugfix release instead.

vladucu:
I think we can likely follow Ember suite and port this to the next 1.x+ version if anybody needs it...no biggie

vladucu:
but yeah, a test would be nice

tomnez:
@vladucu added a test and resolved conflicts, will merge once it passes 🕺

PR #241, opened 2018-10-26
Remove click event object from breadcrumb action args

### LINKED ISSUE #240: Breadcrumb onAction handler gets passed click event object
Switching an app over to use the new breadcrumb implementation, I found that the click event from the `polaris-breadcrumb` `button` element's `onclick` handler was being passed to the outside world on action invocation, which is messy and in my case, caused an error in the action being invoked.

andrewpye:
Fixes #240 by swallowing the click event object when invoking breadcrumb actions.

### DIFF
--- a/addon/templates/components/polaris-breadcrumbs.hbs
+++ b/addon/templates/components/polaris-breadcrumbs.hbs
@@ -15,8 +15,8 @@
   <button
     class="Polaris-Breadcrumbs__Breadcrumb"
     type="button"
-    onclick={{action "fireAction" breadcrumb.onAction}}
     onmouseup={{handleMouseUpByBlurring}}
+    {{action breadcrumb.onAction}}
   >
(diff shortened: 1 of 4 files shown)


addon/templates/components/polaris-breadcrumbs.hbs
```diff
@@ -15,7 +15,7 @@
   <button
     class="Polaris-Breadcrumbs__Breadcrumb"
     type="button"
-    onclick={{action breadcrumb.onAction}}
+    onclick={{action "fireAction" breadcrumb.onAction}}
```
vladucu: would it be the same result just doing `<button {{action breadcrumb.onAction}}...`?

andrewpye: Yep! 👍 Going that way in the interest of cleaner code over matching the Polaris source exactly

Issue #161, opened 2018-07-31
Changes to underlying data don't trigger `polaris-data-table` cell size recalculation


andrewpye:
If a `polaris-data-table` is being used to render data and e.g. a new item is added or the content of a cell is updated, the code to set row heights etc. doesn't run. This can leave table row heights too big/small for their content.

We might be able to handle this by bubbling an action from `polaris-data-table/cell`'s `didRender` hook up to the `polaris-data-table` component and triggering the height recalculation from there.

vladucu:
is this a bug in our implementation or in Shopify's implementation too?

andrewpye:
Actually [looks like it](https://codesandbox.io/s/189y72pl2j), yep - opening an issue on Shopify's repo 👍

vladucu:
@tomnez while cleaning up, would you mind quick checking this one too pls?

PR #243, opened 2018-10-26
Fix data table height not calculating correctly

tomnez:
### Overview

Cell heights weren't being recalculated when underlying data changed (https://github.com/smile-io/ember-polaris/issues/161)

The recalculation issue is [fixed in Polaris](https://github.com/Shopify/polaris-react/blob/master/src/components/DataTable/DataTable.tsx#L108-L113) by recalculating height when the `DataTable` component updates.

### On hold

~~This fix doesn't appear in the React version until **v3.0.0-beta.11**, so we should hold off on merging until we have a 3.0 version of ember-polaris just to keep our versions in sync.~~ (we'll ship now and add a comment)

### Demo

heights

### DIFF
--- a/addon/components/polaris-data-table.js
+++ b/addon/components/polaris-data-table.js
@@ -482,6 +482,12 @@ export default Component.extend(
       this.addEventHandlers();
     },
 
+    didUpdateAttrs() {
+      this._super(...arguments);
+
+      this.handleResize();
+    },
+
     actions: {
       navigateTable(direction) {
         let {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - [#241](https://github.com/smile-io/ember-polaris/pull/241) [FIX] Remove click event from `polaris-breadcrumb` action invocation arguments
 - [#237](https://github.com/smile-io/ember-polaris/pull/237) [FIX] Prevent event bubbling on `polaris-action-list/item`'s `onAction` method
 - [#244](https://github.com/smile-io/ember-polaris/pull/244) [FIX] Fix bug where `polaris-data-table` cell heights weren't calculating correctly
+- [#237](https://github.com/smile-io/ember-polaris/pull/243) [FIX] Update table cell heights when attrs are updated in `polaris-data-table`
 
 ### v1.7.8 (October 10, 2018)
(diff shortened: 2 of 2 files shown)


andrewpye:
>This isn't in version v2.11.0 so I'm not sure if we should repoint this to master or not include it at all yet until we move up to a later version of Polaris where they include it too

My gut says wait until it lands in React world so we're parallel with them as far as possible. We can leave this branch here and if anywhere needs the fix, they can just reference this branch.

>The React version gets away with using childNodes because they use row.childNodes as NodeListOf<HTMLElement>

Good catch, I never stopped to think about why `childNodes` worked for the React version but we had to use `children` 👍

addon/components/polaris-data-table.js
```diff
@@ -482,6 +482,12 @@ export default Component.extend(
       this.addEventHandlers();
     },
 
+    didUpdateAttrs() {
```
andrewpye: I wonder if we're better off using `didRender` rather than `didUpdateAttrs` - reason being that we could potentially be passing objects to represent content (e.g. a hash of `componentName` and `props`); if the content changes but the object reference stays the same, I think `didUpdateAttrs` won't trigger, but `didRender` will.

tomnez: Ohh good call. I'll update 👍

tomnez:
@andrewpye updated to use `didRender` and also took out the `childNodes` fix which I will open in a separate PR.

andrewpye:
>I don't think we should be 1:1 with any issues in the React implementation

I _do_ see your point, but at the same time, fixing bugs in parallel with the React implementation does make issue tracking/resolution more straightforward in Ember-land...

vladucu:
> I do see your point, but at the same time, fixing bugs in parallel with the React implementation does make issue tracking/resolution more straightforward in Ember-land...

we should then maybe add a comment, but definitely not wait to update to matching react version if it was already fixed in React in a newer version...especially if it's a straight-forward fix

tomnez:
Ok, we'll get this in now after the build completes instead of waiting. I'll add a comment 👍

vladucu:
thx boizz
🚀

tomnez:
@andrewpye reverted back to using `didUpdateAttrs`, not sure what's going on in ember test land, but  using `didRender` seemed to timeout all the tests for `polaris-data-table` 🤔I guess can revisit if `didUpdateAttrs` turns out to be not good enough.

Issue #242, opened 2018-10-26
polaris-select considers wrong item selected on initialisation


andrewpye:
Noticed that on first initialisation, `polaris-select` always seems to consider the first option selected irrespective of the `value` passed in; this doesn't seem to be an issue in the React implementation.

In the GIF below, the selected value is `stripe`, but the control shows the checkmark next to the first item (`shopify`), which prevents that first option being clickable until another option has been clicked:

![polaris-select-bug](https://user-images.githubusercontent.com/5737342/47569579-9f610180-d93c-11e8-9d37-80f1c7b8d48e.gif)

tomnez:
@andrewpye I looked into this and can't really see any big difference in implementation other than us using `value={{selectedOption}}` instead of `value={value}` to set the `value` attr of the `select` tag. But changing to `value={{value}}` doesn't make a difference either.

I tried waiting until `selectedOption` was set to render the select field, but that didn't help (thought maybe it could be a weird render issue with setting value after render or something). The only thing that seemed to work for me was explicitly adding a `selected` attr to the `option` tags:

```hbs
<option
  selected={{eq selectedOption opt.label}}
  value={{opt.value}}
  disabled={{opt.disabled}}
  hidden={{opt.hidden}}
>
```

but that goes against the React implementation so I'm not sure if we want to do that. Thoughts?

andrewpye:
@tomnez thanks for investigating! I think we should definitely update to use `value={{value}}` instead of `value={{selectedOption}}` (that just seems like a small bug that slipped through). I've tested something like your suggested fix of passing `selected` explicitly on the `option` element and it seems to solve the problem. I'd say let's roll with that - I'd rather differ slightly from the React code and have the same functionality/lack of bugs, than match their code word-for-word 😉 The only thing I'd suggest would be to pass `selected={{eq opt.value value}}` since it seems slightly more prudent to match on the underlying values rather than on labels. I'll let you put a PR up to claim the glory for this one! 💪🔥

andrewpye:
Closed in #245

PR #245, opened 2018-10-29
Bugfix polaris-select wrong selected option on init

tomnez:
Fixes https://github.com/smile-io/ember-polaris/issues/242

### DIFF
--- a/addon/templates/components/polaris-select.hbs
+++ b/addon/templates/components/polaris-select.hbs
@@ -11,7 +11,7 @@
       class="Polaris-Select__Input"
       id={{_id}}
       name={{name}}
-      value={{selectedOption}}
+      value={{value}}
       disabled={{disabled}}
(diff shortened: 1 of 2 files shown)


vladucu:
@tomnez this looks sane to me, just curious what's the actual issue? is there no selected option on init? or a wrong one?

tomnez:
Yeah on init it will _display_ the correctly chosen option, but if you then click the select to see other options, the browser's "checkmark" will be displayed on the first option and not the chosen option.

You can see a gif Andy captured in the issue - https://github.com/smile-io/ember-polaris/issues/242

andrewpye:
@tomnez just realised that this needs a changelog entry. I'm also wondering if it's possible to add a test for this - I think that before making these changes, the `select` element won't have a `value` set on initialisation, but after this fix it will, perhaps we can leverage that to ensure we don't see this issue come up again in the future?

PR #251, opened 2018-11-09
Make allowRange on polaris-date-picker writeable

andrewpye:
The React implementation of `DatePicker` has `allowRange` as a [public prop](https://github.com/Shopify/polaris-react/blob/master/src/components/DatePicker/DatePicker.tsx#L36). We need to make it writeable so that we can enforce selecting single dates.


vladucu:
💪

andrewpye:
Gonna leave this hanging for now until we can merge it into `master`

PR #254, opened 2018-11-26
Add checkable-button component for resource-list

tomnez:
Adds new `resource-list/checkable-button` subcomponent needed for the main `resource-list` component.

### DIFF
--- a/addon/components/polaris-resource-list/checkable-button.js
+++ b/addon/components/polaris-resource-list/checkable-button.js
@@ -1,0 +1,106 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import layout from '../../templates/components/polaris-resource-list/checkable-button';
+
+export default Component.extend({
+  layout,
+
+  /**
+   * @type {String}
+   * @default null
+   * @property accessibilityLabel
+   */
+  accessibilityLabel: null,
+
+  /**
+   * @type {String}
+   * @default ''
+   * @property label
+   */
+  label: '',
+
+  /**
+   * @type {Boolean|String}
+   * @default null
+   * @property selected
+   */
+  selected: null,
+
+  /**
+   * @type {Boolean}
+   * @default false
+   * @property selectMode
+   */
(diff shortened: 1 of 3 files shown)


addon/components/polaris-resource-list/checkable-button.js
```diff
@@ -0,0 +1,106 @@
...
+   * @property label
+   */
+  label: '',
+
+  /**
+   * @type {Boolean|String}
+   * @default null
+   * @property selected
+   */
+  selected: null,
```
tomnez: Do we need to handle this similar to how we handle a selected value in the `polaris-checkbox`? It doesn't look like the `ResourceList` [component](https://github.com/Shopify/polaris-react/blob/master/src/components/ResourceList/ResourceList.tsx#L427) passes anything down into `CheckableButton` for `selected`, so I wonder if our implementation should just let the `polaris-checkbox` in the template handle setting it's value and this `selected` attribute here will get its value based on whatever gets passed back up from the `polaris-checkbox` (?) It looks like this is only used for setting the `--selected` class.

andrewpye: `ResourceList` itself doesn't seem to pass `selected`, but `ResourceList/BulkActions` [does](https://github.com/Shopify/polaris-react/blob/master/src/components/ResourceList/components/BulkActions/BulkActions.tsx#L327), if that changes anything...

addon/templates/components/polaris-resource-list/checkable-button.hbs
```diff
@@ -0,0 +1,11 @@
+<div class={{checkableButtonClasses}} onclick={{action onToggleAll}}>
```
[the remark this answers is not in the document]
tomnez: Ahh yeah forgot the `tagName: ''`, derp. But yeah that sounds good, I forgot about `mapEventToAction` 👍

tomnez:
@andrewpye updated.

PR #256, opened 2018-11-28
Move ember-lifeline to deps

sivakumar-kailasam:
w/o this a fresh installation of the addon fails.


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # ember-polaris Changelog
 
+### v2.0.4 (November 28, 2018)
```
andrewpye: I'd normally put this as `Unreleased` and update it on master when we cut a release, so that a) we can group changes together under releases more easily, and b) there's no assumption of what release a PR will go in and when it'll be published.

PR #257, opened 2018-11-28
Add resource list date selector

andrewpye:
Adds an implementation of the React [`ResourceList/FilterControl/DateSelector`](https://github.com/Shopify/polaris-react/tree/master/src/components/ResourceList/components/FilterControl/components/DateSelector) component. **NOTE:** some tests are currently skipped owing to either having a timing issue I haven't been able to resolve, or not having figured out a way to port that particular test to Ember-land 😬


andrewpye:
Turns out that the timezone issue I'm seeing in some tests is also present in the React implementation (see https://github.com/Shopify/polaris-react/issues/688), so I'll leave these as they are for now and we can address this later when a fix is available 👍

addon/templates/components/polaris-resource-list/filter-control/date-selector.hbs
```diff
@@ -0,0 +1,32 @@
+{{polaris-select
+  label="Select a value"
+  labelHidden=true
+  placeholder="YYYY-MM-DD"
```
vladucu: should this placeholder be something like [here](https://github.com/Shopify/polaris-react/blob/master/src/locales/en.json#L148)

tests/integration/components/polaris-resource-list/filter-control/date-selector-test.js
```diff
@@ -0,0 +1,617 @@
...
+  if (!options) {
+    return [];
+  }
+
+  return options.map((option) => {
+    return typeof option === 'string' ? option : option.value;
+  });
+}
+
+async function triggerChangeEventWithValue(selector, value) {
```
vladucu: not 💯 but you might be able to just use [`fillIn`](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#fillin) instead of this custom method

andrewpye: I tried `fillIn` but couldn't get it to work... 🤷‍♂️

addon/templates/components/polaris-resource-list/filter-control/date-selector.hbs
andrewpye: Good catch, thanks! 🦅👀

PR #258, opened 2018-11-29
Add resource list filter value selector

andrewpye:
Adds an implementation of the React [`ResourceList/FilterControl/FilterValueSelector`](https://github.com/Shopify/polaris-react/tree/master/src/components/ResourceList/components/FilterControl/components/FilterValueSelector) component.


addon/components/polaris-text-field.js
```diff
@@ -561,11 +561,6 @@ export default Component.extend(
 
       let { id, type } = this.getProperties('id', 'type');
 
-      assert(
-        `ember-polaris::polaris-text-field - ${type} is not a valid type.`,
-        allowedTypes.indexOf(type) > -1
-      );
-
```
andrewpye: @tomnez not sure if this was needed but it was causing a test error on the new tests I've added here so I ripped it out 😬

tomnez: Hm I guess we can leave it up to users to just use `polaris-text-field` responsibly if it's causing issues 🤷‍♂️ 

Alternatively, are there types you could add into `allowedTypes` that would stop the test from erroring out?

andrewpye: I don't hugely want to have to add `undefined` to the whitelist... 😬

tomnez: Haha oh, I was thinking it might be a real value (didn't look through the whole PR yet) so yeah, lets not do that 😝

addon/templates/components/polaris-resource-list/filter-control/filter-value-selector.hbs
```diff
@@ -0,0 +1,41 @@
+{{#if (eq filter.type FilterType.Select)}}
+  {{#polaris-stack vertical=true as |stack|}}
+    {{#if showOperatorOptions}}
+      {{#stack.item}}
+        {{render-content operatorOptionsMarkup}}
+      {{/stack.item}}
+    {{/if}}
+    {{polaris-select
```
vladucu: nit: some spacing before this maybe

addon/templates/components/polaris-resource-list/filter-control/filter-value-selector.hbs
```diff
@@ -0,0 +1,41 @@
...
+    }}
+  {{/polaris-stack}}
+{{else if (eq filter.type FilterType.TextField)}}
+  {{#polaris-stack vertical=true as |stack|}}
+    {{#if showOperatorOptions}}
+      {{#stack.item}}
+        {{render-content operatorOptionsMarkup}}
+      {{/stack.item}}
+    {{/if}}
+    {{polaris-text-field
```
vladucu: nit: some spacing before this maybe

addon/templates/components/polaris-resource-list/filter-control/filter-value-selector.hbs
andrewpye: Normally I would, but decided to match the React source here 🤷‍♂️

PR #259, opened 2018-11-29
Add polaris-resource-list/bulk-actions component

tomnez:
### Overview

Adds the `polaris-resource-list/bulk-actions` component needed for `polaris-resource-list` 🕺 

- Resource list [demo](https://polaris.shopify.com/components/lists-and-tables/resource-list)
- [BulkActions](https://github.com/Shopify/polaris-react/blob/master/src/components/ResourceList/components/BulkActions/BulkActions.tsx) component on github

**Note:** The react implementation has an attribute named `actions` which we obviously can't use as an attribute name, so in this implementation I renamed it to `passedActions`.

### Demo

The demo has `selectMode` hard-coded to `true`, so it looks a bit incomplete since it's outside of its parent component.

bulk-actions


addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,535 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { throttle } from '@ember/runloop';
```
andrewpye: Would we maybe be better using `throttleTask` from `ember-lifeline` for this?

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,535 @@
...
+   * Disables bulk actions
+   *
+   * @type {Boolean}
+   * @default null
+   * @property disabled
+   */
+  disabled: null,
+
+  /**
+   * @private
```
andrewpye: I haven't been tagging internal/private props like this as the linter considers them private by default and it just adds clutter - thoughts?

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,535 @@
...
+
+  hasActions: computed('promotedActions', 'actionsCollection', function() {
+    let { promotedActions, actionsCollection } = this.getProperties(
+      'promotedActions',
+      'actionsCollection'
+    );
+
+    return Boolean(
+      (promotedActions && promotedActions.length > 0) ||
+        (actionsCollection && actionsCollection.length > 0)
```
andrewpye: Shouldn't this line be at the same indentation as the one above it?

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,535 @@
...
+  ),
+
+  /**
+   * Callback when selectable state of list is changed
+   *
+   * @type {Function}
+   * @default noop
+   * @property onSelectModeToggle
+   */
+  onSelectModeToggle(/* selectMode **/) {},
```
andrewpye: This and `onToggleAll` below should be tagged as `@public` and moved up beneath the other public props

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,535 @@
...
+    let validList = actionsCollection.filter((action) => {
+      return !action.items;
+    });
+
+    return actionsCollection.length === validList.length;
+  },
+
+  setContainerNode() {
+    this.set('containerNode', this.element);
+  },
```
andrewpye: I don't see any real point to doing this - if you want to set `containerNode` to `this.element`, maybe use `alias`? Or we could just use `this.element` everywhere the React code uses `this.containerNode`, aliasing it seems a bit pointless 🤷‍♂️

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,535 @@
...
+      '.Polaris-ResourceList-BulkActions__Group--largeScreen'
+    );
+
+    if (moreActionsEl) {
+      this.setMoreActionsNode(moreActionsEl);
+    }
+
+    if (largeScreenButtonsEl) {
+      this.setLargeScreenButtonsNode(largeScreenButtonsEl);
+    }
```
andrewpye: This seems like a bit of a clunky approach to imitating React's `ref` behaviour - in the past we've handled this by defining [CPs](https://github.com/smile-io/ember-polaris/blob/master/addon/mixins/components/svg-handling.js#L50) that look up the element, do you think that approach would be cleaner here?

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,535 @@
...
+    });
+
+    return actionsCollection.length === validList.length;
+  },
+
+  setContainerNode() {
+    this.set('containerNode', this.element);
+  },
+
+  setMoreActionsNode(node) {
```
andrewpye: See comment on `didRender` re. this and `setLargeScreenButtonsNode`

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,535 @@
...
+  },
+
+  setLargeScreenButtonsNode(node) {
+    this.set('largeScreenButtonsNode', node);
+  },
+
+  handleResize() {
+    throttle(
+      this,
+      function() {
```
andrewpye: Just want to check that this throttling is working... I know that things like `scheduleOnce` can lose their "once" behaviour if you pass in an anonymous function rather than a method on the object, for example. In this case, each time `handleResize` gets called, `throttle` will be called with a _new_ anonymous function passed in as the callback each time; if `throttle` is using the function pointer to determine if it should run the callback yet, then the throttling behaviour will be lost. I mean, if the throttle implementation is something like this:

```js
throttle(target, callback, ...) {
  if (target === lastTarget && callback === lastCallback) {
    // do the throttling
  } else {
    // start throttle queue for callback
  }
}
```

then you have to pass the _same_ function instance each time you call it for throttling to work. Hope that makes sense 😅

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,535 @@
...
+    );
+  },
+
+  didInsertElement() {
+    this._super(...arguments);
+
+    let promotedActions = this.get('promotedActions');
+
+    if (promotedActions && promotedActions.length > MAX_PROMOTED_ACTIONS) {
+      warn(
```
andrewpye: `warn` [needs](https://github.com/emberjs/ember.js/blob/v3.5.1/packages/%40ember/debug/lib/warn.ts#L102) to be given an `id` for the warning otherwise this will log a deprecation warning. We're already doing this in e.g. [`polaris-popover`](https://github.com/smile-io/ember-polaris/blob/master/addon/components/polaris-popover.js#L167) so just follow suit from that 👍

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,535 @@
...
+
+    this.setContainerNode();
+
+    let largeScreenGroupNode = this.element.querySelector(
+      '.Polaris-ResourceList-BulkActions__Group--largeScreen'
+    );
+
+    if (largeScreenGroupNode) {
+      this.addEventListener(largeScreenGroupNode, 'resize', this.handleResize);
+    }
```
andrewpye: Can we maybe move this event listener setup into its own method for clarity?

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,535 @@
...
+      });
+    }
+  },
+
+  actions: {
+    toggleSmallScreenPopover() {
+      this.set(
+        'smallScreenPopoverVisible',
+        !this.get('smallScreenPopoverVisible')
+      );
```
andrewpye: Currently `polaris-popover` behaves differently to the React implementation in that it's opened/closed using actions and the `ember-basic-dropdown` trigger, so these popover properties/actions won't have the same effect as the React component. Can you look at the React implementation's behaviour and figure out a way to replicate it using `polaris-popover` if possible (assuming they behave differently... if they behave the same already then look at removing any popover-related code from here that doesn't have any effect) 👍

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,535 @@
...
+      this.set(
+        'largeScreenPopoverVisible',
+        !this.get('largeScreenPopoverVisible')
+      );
+    },
+
+    setSelectMode(val) {
+      let onSelectModeToggle = this.get('onSelectModeToggle');
+
+      if (onSelectModeToggle) {
```
andrewpye: Since we default this to a no-op, there isn't really any need for `setSelectMode`, we could just pass `(action onSelectModeToggle)` in the template 🤷‍♂️

addon/templates/components/polaris-resource-list/bulk-actions.hbs
```diff
@@ -0,0 +1,139 @@
+<div
+  class="Polaris-ResourceList-BulkActions__Group
+    Polaris-ResourceList-BulkActions__Group--smallScreen
+    {{if selectMode 'Polaris-ResourceList-BulkActions__Group--entered' 'Polaris-ResourceList-BulkActions__Group--exited'}}"
```
andrewpye: Nit: can you multi-line this guy please? Also the single quotes seem to be causing a template lint error 😬

addon/templates/components/polaris-resource-list/bulk-actions.hbs
```diff
@@ -0,0 +1,139 @@
...
+      selected=selected
+      selectMode=selectMode
+      measuring=measuring
+      disabled=disabled
+      onToggleAll=(action onToggleAll)
+    }}
+
+    {{#if hasActions}}
+      <div class="Polaris-ResourceList-BulkActions__Popover">
+        {{!-- TODO: set `active` to smallScreenPopoverVisible once implemented --}}
```
andrewpye: See earlier comments about popover differences between Ember and React land 😉

addon/templates/components/polaris-resource-list/bulk-actions.hbs
```diff
@@ -0,0 +1,139 @@
...
+      <div class="Polaris-ResourceList-BulkActions__Popover">
+        {{!-- TODO: set `active` to largeScreenPopoverVisible once implemented --}}
+        {{#polaris-popover
+          onClose=(action "toggleLargeScreenPopover")
+          as |popover|
+        }}
+          {{#popover.activator}}
+            {{polaris-resource-list/bulk-actions/bulk-action-button
+              content=activatorLabel
+              disabled=disabled
```
andrewpye: Check the behaviour of the popover here when `disabled` is truthy - I found the other day that despite setting the popover activator button to be disabled, the activator itself still worked, so the popover could still be triggered even though the button itself was disabled 😬 I worked around it by doing this:

```hbs
{{#component (if disabled
  (component "wrapper-element" tagName="")
  popover.activator
)}}
  {{polaris-button
    ...
    disabled=disabled
    ...
  }}
{{/component}}
```

to swap out the popover activator when the button's disabled, but maybe there's a better way? 🤔

addon/templates/components/polaris-resource-list/bulk-actions.hbs
```diff
@@ -0,0 +1,139 @@
...
+              sections=combinedActions
+              onActionAnyItem=(action "toggleLargeScreenPopover")
+            }}
+          {{/popover.content}}
+        {{/polaris-popover}}
+      </div>
+    {{/if}}
+  </div>
+
+  {{#if (or paginatedSelectAllAction (and paginatedSelectAllText paginatedSelectAllAction))}}
```
andrewpye: Multi-line?

tests/integration/components/polaris-resource-list/bulk-actions-test.js
```diff
@@ -0,0 +1,12 @@
...
+// import { render } from '@ember/test-helpers';
+// import hbs from 'htmlbars-inline-precompile';
+
+// module('Integration | Component | polaris-resource-list/bulk-actions', function(hooks) {
+//   setupRenderingTest(hooks);
+
+//   test('it renders', async function(assert) {
+
+//   });
+// });
```
andrewpye: Port the tests across from the React implementation pls 😅

addon/components/polaris-resource-list/bulk-actions.js
tomnez: Yeah I'm good with removing 👍

addon/components/polaris-resource-list/bulk-actions.js
tomnez: I grabbed this right from the react implementation where they have an indent 🤷‍♂️ I'll fix it up here tho.

addon/components/polaris-resource-list/bulk-actions.js
tomnez: Ahh I left out the `@public` 🤦‍♂️ Was wondering why prettier was mad at me.

addon/components/polaris-resource-list/bulk-actions.js
tomnez: Ah I hadn't thought of that, yeah lets just use `this.element`

addon/components/polaris-resource-list/bulk-actions.js
tomnez: Ah right - I can refactor to just use `popover.close` in the `onActionAnyItem` call of the popover, but do you have any thoughts on how we can close the popover in the `resize` handler?

The react implementation just sets `smallScreenPopoverVisible` and `smallScreenPopoverVisible` to `false` on `resize` but we aren't able to take advantage of this and I'm not sure of a good way to pass the currently active `popover` object into the component's methods 🤔

tests/integration/components/polaris-resource-list/bulk-actions-test.js
tomnez: screen shot 2018-12-03 at 4 53 24 pm

addon/components/polaris-resource-list/bulk-actions.js
andrewpye: Hmm... maybe it's time we ported the [`EventListener`](https://github.com/Shopify/polaris-react/blob/master/src/components/EventListener/EventListener.tsx) component and use that - that way we could maybe pass `popover`/`popover.close` into the action handler, either directly (e.g. `onAction=(action popover.close)`) or indirectly (`onAction=(action "someAction" popover`, and then `someAction` does whatever it needs to do plus calls `popover.close()`) 🤔

addon/components/polaris-resource-list/bulk-actions.js
andrewpye: This is unused now so should be removed 👍

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -0,0 +1,460 @@
...
+    setContainerWidth() {
+      let containerWidth = this.element.getBoundingClientRect().width;
+
+      if (containerWidth > 0) {
+        this.set('containerWidth', containerWidth);
+      }
+    },
+
+    handleResize() {
+      this.throttleTask('setContainerWidth', 50);
```
andrewpye: 👍

addon/templates/components/polaris-resource-list/bulk-actions.hbs
andrewpye: Yeah it's pretty icky 😬

PR #260, opened 2018-11-30
Add resource list filter creator component

andrewpye:
Adds an implementation of the React [`ResourceList/FilterControl/FilterCreator`](https://github.com/Shopify/polaris-react/tree/master/src/components/ResourceList/components/FilterControl/components/FilterCreator) component. **NOTE:** one of the ported tests is currently skipped owing to not being able to figure out why it won't pass 😅


addon/components/component-proxy.js
```diff
@@ -87,6 +87,7 @@ export default Component.extend({
 
     let propsToUnset = oldPropNames.reduce((propsToUnset, propName) => {
       propsToUnset[propName] = undefined;
+      return propsToUnset;
```
andrewpye: No idea how this didn't blow up on us before, showed up in one of the new tests I added for this PR 😅

addon/components/polaris-resource-list/filter-control/filter-creator.js
```diff
@@ -0,0 +1,156 @@
...
+    );
+
+    if (!onAddFilter || !this.get('canAddFilter') || !selectedFilterKey) {
+      return;
+    }
+
+    onAddFilter({
+      key: selectedFilterKey,
+      value: this.get('selectedFilterValue') || '',
+    });
```
tomnez: nit: I know this matches the react file but maybe add a space here?

tests/integration/components/polaris-resource-list/filter-control/filter-creator-test.js
```diff
@@ -0,0 +1,468 @@
...
+          disabled=disabled
+          onAddFilter=(action (mut wasOnAddFilterCalled) true)
+        }}
+      `);
+
+      await activatePopover();
+      await selectFilterKey(filters[0].key);
+      await selectFilterValue('Bundle');
+
+      assert.dom('[data-test-select="filter"]').hasAnyValue();
```
tomnez: I think this is borking here because `hasAnyValue` eventually makes its way into the `hasValue` assertion under the hood, where it grabs the target using [`element as HTMLInputElement`](https://github.com/simplabs/qunit-dom/blob/master/lib/assertions.ts#L696). In this test we're grabbing `'[data-test-select="filter"]'` which is an `HTMLSelectElement` type.

So something like

```javascript
let filterSelectValue = find('[data-test-select="filter"]').value;

assert.ok(filterSelectValue);
```

would probably work here instead?

tomnez: (and then same for the next assertion)

tomnez: aw bummer 😞 ok then.

PR #263, opened 2018-12-10
Add polaris-empty-search-result component

tomnez:
Adding [EmptySearchResult](https://github.com/Shopify/polaris-react/tree/master/src/components/EmptySearchResult) component 🕺


addon/templates/components/polaris-empty-search-result.hbs
```diff
@@ -0,0 +1,25 @@
+{{#polaris-stack
+  alignment="center"
+  vertical=true
+}}
+  {{#if withIllustration}}
+    {{!-- TODO: replace this with polaris-image when implemented. --}}
+    <img
+      src="/illustrations/empty-search.svg"
```
tomnez: Also TODO before merging, I image this path will have to be updated. Pulling this branch into a test app now to check.

addon/templates/components/polaris-empty-search-result.hbs
```diff
@@ -0,0 +1,25 @@
+{{#polaris-stack
+  alignment="center"
+  vertical=true
+}}
+  {{#if withIllustration}}
+    {{!-- TODO: replace this with polaris-image when implemented. --}}
+    <img
```
andrewpye: This should be wrapped in an explicit `stack.item` to avoid a potential [glimmer error](https://github.com/smile-io/ember-polaris/pull/159) if `withIllustration` changes from truthy to falsy

addon/templates/components/polaris-empty-search-result.hbs
```diff
@@ -0,0 +1,25 @@
...
+      src="empty-search.svg"
+      alt="Empty search results"
+      class="Polaris-EmptySearchResult__Image"
+    >
```
andrewpye: Any reason for not having the [`draggable`](https://github.com/Shopify/polaris-react/blob/master/src/components/EmptySearchResult/EmptySearchResult.tsx#L41) attribute on this?

addon/templates/components/polaris-empty-search-result.hbs
```diff
@@ -0,0 +1,25 @@
...
+  {{/if}}
+
+  {{polaris-display-text
+    size="small"
+    text=title
+  }}
+
+  {{#if description}}
+    {{polaris-text-style
```
andrewpye: Again, add the explicit `stack.item` here pls

addon/templates/components/polaris-empty-search-result.hbs
```diff
@@ -0,0 +1,25 @@
...
+      variation="subdued"
+      text=description
```
andrewpye: The description text should be [wrapped](https://github.com/Shopify/polaris-react/blob/master/src/components/EmptySearchResult/EmptySearchResult.tsx#L34) in a `p` element 😉

addon/templates/components/polaris-empty-search-result.hbs
tomnez: Nope just missed it 😳

addon/templates/components/polaris-empty-search-result.hbs
```diff
@@ -1,14 +1,18 @@
...
-    {{!-- TODO: replace this with polaris-image when implemented. --}}
-    <img
-      src="empty-search.svg"
-      alt="Empty search results"
-      class="Polaris-EmptySearchResult__Image"
-    >
+    {{#stack.item}}
+      {{!-- TODO: replace this with polaris-image when implemented. --}}
+      <img
+        src="empty-search.svg"
```
andrewpye: Does this Just Work™️ if you use this in a real host app, or does it need the image file to be copied across?

tomnez: The image is added in this PR 😄

tomnez: When the main component is done we should (obviously) pull the branch into an app and make sure the path is correct but I _think_ this should be ok.

andrewpye: Yeah agreed - there's an image in the main component too so we should check both 😉

PR #264, opened 2018-12-12
Finish tests for polaris-resource-list/item component

tomnez:


addon/templates/components/polaris-resource-list/item.hbs
```diff
@@ -76,7 +76,7 @@
         data-test-id=(unless persistActions "shortcut-actions")
         as |buttonGroup|
       }}
-        {{#each shortcutActions as shortcutAction}}
+        {{#each shortcutActions as |shortcutAction|}}
```
tomnez: 🙈

tests/integration/components/polaris-resource-list/item-test.js
```diff
@@ -315,8 +318,31 @@ module('Integration | Component | polaris-resource-list/item', function(hooks) {
           .dom('[data-test-checkbox-input]')
           .hasAttribute('data-test-checkbox-input-checked');
       });
+    }
+  );
 
-      test("renders a disabled checked Checkbox if 'loading' context is true", async function(assert) {
```
tomnez: This test was nested under the `selectMode` module, but in `polaris-react` it looks like it's nested under the `media` module so it hasn't been removed, just moved down into `media` FYI.

tests/integration/components/polaris-resource-list/item-test.js
```diff
@@ -325,10 +351,89 @@ module('Integration | Component | polaris-resource-list/item', function(hooks) {
...
+            {{polaris-resource-list/item
+              persistActions=true
+              id=itemId
+              url=url
+              shortcutActions=shortcutActions
+            }}
+          {{/polaris-resource-list/provider}}
+        `);
+
+        assert.dom('.Polaris-ResourceList-Item__Disclosure').exists();
```
tomnez: `polaris-react` tests for a [`ButtonGroup`](https://github.com/Shopify/polaris-react/blob/master/src/components/ResourceList/components/Item/tests/Item.test.tsx#L372) here, but it looks like our template renders a `polaris-button-group` either way, so I'm checking for `Disclosure` class instead since that's only rendered if `persistActions` is true.

andrewpye: Fair point, can you leave this as a comment so that we know why we're differing from the original test here? 👍

tests/integration/components/polaris-resource-list/item-test.js
andrewpye: Yeah looks like I messed up the test layout somehow when I first did these 🤦‍♂️

addon/templates/components/polaris-resource-list/item.hbs
andrewpye: 🤦‍♂️

tests/integration/components/polaris-resource-list/item-test.js
andrewpye: Wait wat... that test has moved since I pulled the React code to port it 😅 Seems like an accident as well, what are your thoughts on keeping it here where it belongs irrespective of what's in the React code?

tomnez: Hm good point. Also that's weird that it moved since you started haha. I'll put it back.

PR #265, opened 2018-12-13
Rewrite main resource list component

andrewpye:
Completely overhauls the old, broken, outdated `polaris-resource-list` with the shiny new version. Sorry for the somewhat beastly PR 😅


addon/components/polaris-resource-list.js
```diff
@@ -1,34 +1,797 @@
...
+
+      if (this.get('loading')) {
+        this.setLoadingPosition();
+      }
+    },
+
+    didRender() {
+      this._super(...arguments);
+
+      /*
```
tomnez: nit: comment indentation is off 😛

addon/components/polaris-resource-list.js
```diff
@@ -1,34 +1,797 @@
...
+     * The React implementation currently caches `listNode` on first render,
+     * which leads to some buggy behaviour around whether the header gets
+     * rendered or not (see https://github.com/Shopify/polaris-react/issues/735).
+     * This is how I've chosen to fix the issue here, but we'll keep an eye
+     * on if/how the React implementation chooses to fix it and can maybe update
+     * to match their fix later if it seems better in some way.
+     */
+      this.setListNode();
```
tomnez: All good with me since we're targeting with `guidFor` 👍

addon/components/polaris-resource-list.js
```diff
@@ -1,34 +1,797 @@
...
+      }
+    },
+
+    handleSelectAllItemsInStore() {
+      let {
+        onSelectionChange,
+        selectedItems,
+        items,
+        idForItem,
+      } = this.getProperties('onSelectionChange', 'selectedItems', 'items');
```
tomnez: Missing `idForItem` in your `getProperties` here.

addon/components/polaris-resource-list.js
```diff
@@ -1,34 +1,797 @@
...
+      this.setProperties({
+        previousSelectedItems: selectedItems,
+        previousLoading: loading,
+      });
+    },
+
+    didInsertElement() {
+      this._super(...arguments);
+
+      this.addEventListener(window, 'resize', this.handleResize);
```
tomnez: It looks like [react](https://github.com/Shopify/polaris-react/blob/v2.11.0/src/components/ResourceList/ResourceList.tsx#L353) only has this listener in the DOM when `selectable` is true, would it be worth it to add something like:

```javascript
debouncedHandleResize() {
  if (!this.get('selectable') {
    return;
  }
 // ...
}
```

at the beginning of `debouncedHandleResize`? I'm fine with leaving it though, doesn't seem to be any harm if this gets called when `selectable` is false.

addon/templates/components/polaris-resource-list.hbs
```diff
@@ -1,5 +1,145 @@
...
+        {{/polaris-sticky}}
+      </div>
+    {{/if}}
+
+    {{#if itemsExist}}
+      {{!--
+        NOTE: Polaris-ResourceList--disabledPointerEvents doesn't exist at this point,
+        but Polaris-ResourceList__DisabledPointerEvents does. The former is probably
+        what was intended in the React source, so matching that for now as the class
+        doesn't seem to be necessary for the list's `loading` behaviour to work.
```
tomnez: 👍 🤷‍♂️

addon/components/polaris-resource-list.js
```diff
@@ -1,34 +1,797 @@
...
+              )} in your store`;
+
+        return {
+          content: actionText,
+          onAction: () => this.handleSelectAllItemsInStore,
+        };
+      }
+    ).readOnly(),
+
+    emptySearchResultTitle: computed('resourceName.plural', function() {
```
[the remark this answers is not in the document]
andrewpye: Yeah, but their `emptySearchResultText` is a hash of `description` and `title`, I'm using the description inline and only needed to compute the title in the JS file 😉

addon/components/polaris-resource-list.js
```diff
@@ -1,34 +1,797 @@
...
+     * @property listNode
+     * @type {HTMLUListElement}
+     * @default null
+     * @private
+     */
+    listNode: null,
+
+    /**
+     * @property defaultResourceName
+     * @type {Object}
```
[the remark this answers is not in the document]
andrewpye: This is private and gets initialised internally then never updated, so I don't see any real point to documenting it that much 😛

addon/components/polaris-resource-list.js
```diff
@@ -1,34 +1,797 @@
...
+
+    /**
+     * Internal property used to recreate React implementation's
+     * `componentDidUpdate` behaviour.
+     *
+     * @property previousLoading
+     * @type {Boolean}
+     * @private
+     */
+    previousLoading: null,
```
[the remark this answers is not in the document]
andrewpye: Yeah I guess 😛

addon/components/polaris-resource-list.js
andrewpye: Great catch! 🙌

andrewpye:
Added an example page to the dummy app to check the behaviour against the React [examples](https://polaris.shopify.com/components/lists-and-tables/resource-list). Apart from the icon rendering issue in the dummy app, I think this looks pretty good:

resource-list-examples

addon/templates/components/polaris-resource-list/item.hbs
```diff
@@ -88,7 +92,9 @@
               destructive=shortcutAction.destructive
               icon=shortcutAction.icon
               loading=shortcutAction.loading
-              onClick=(action shortcutAction.onAction)
+              onClick=(action
+                (or shortcutAction.onAction (action (mut dummy)))
```
tomnez: @andrewpye why would there be a shortcutAction button without an action?

andrewpye: Could be a `url` action for a link 😉 (I hit this testing the examples in the dummy app)

tomnez: Ohh right.

PR #269, opened 2018-12-20
Remove listener mutating text-field value

vladucu:
Diff better viewed [here](https://github.com/smile-io/ember-polaris/pull/269/files?w=1)


addon/components/polaris-text-field.js
```diff
@@ -37,625 +35,602 @@ const dpl = (num) => (num.toString().split('.')[1] || []).length;
...
+      let numericValue = value ? parseFloat(value) : 0;
+
+      if (isNaN(numericValue)) {
+        return;
+      }
+
+      // Making sure the new value has the same length of decimal places as the
+      // step / value has.
+      let decimalPlaces = Math.max(dpl(numericValue), dpl(step));
+      let newValue = Math.min(max, Math.max(numericValue + steps * step, min));
```
vladucu: not really sure why linter changed these

vladucu:
@tomnez I think I've found the root cause for this not working as expected...we've been using `onchange` instead of `oninput` event....can you give this a try please? (I've tested with the dummy app worked well)

PR #273, opened 2019-01-24
Add option list component

andrewpye:
Adds a port of the React [`OptionList` component](https://polaris.shopify.com/components/lists-and-tables/option-list). This is based on the [source](https://github.com/Shopify/polaris-react/tree/v2.11.0/src/components/OptionList) for the version of Polaris that we're currently targeting in `ember-polaris` (i.e. v2.11.0).


addon/components/polaris-option-list.js
```diff
@@ -0,0 +1,185 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { guidFor } from '@ember/object/internals';
+import layout from '../templates/components/polaris-option-list';
+
+/**
+ * Polaris option list component.
+ * See https://polaris.shopify.com/components/lists-and-tables/option-list
```
tomnez: Do we still need this here since we now have individual docs with the links per component? No biggie though. Just a thought.

andrewpye: This is in the majority of our other components, I don't see any harm having an extra link directly in the source code ¯\\\_(ツ)_/¯

PR #274, opened 2019-01-30
Add support for using a block for render-content

vladucu:
### Overview

This enhances our internal `render-content` allowing us to default to using a specific component when the content is a string.

#### Example
`my-component`
```js
// String | Component
title: 'Some default'
```

```hbs
...
{{#render-content title}}
  {{!-- render the title as sub-heading, by default --}}
  {{polaris-subheading text=title}}
{{/render-content}}
```

### DIFF
--- a/addon/templates/components/render-content.hbs
+++ b/addon/templates/components/render-content.hbs
@@ -3,5 +3,9 @@
 {{else if contentIsComponentDefinition}}
   {{component content}}
 {{else}}
-  {{content}}
+  {{#if hasBlock}}
+    {{yield}}
+  {{else}}
+    {{content}}
+  {{/if}}
 {{/if}}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # ember-polaris Changelog
 
+### Unreleased
+- [#274](https://github.com/smile-io/ember-polaris/pull/274) [ENHANCEMENT] Add support for using a block for `render-content`
+
 ### v3.0.5 (January 28, 2019)
 - [#273](https://github.com/smile-io/ember-polaris/pull/273) [FEATURE] Add `polaris-option-list` component
 
--- a/tests/integration/components/render-content-test.js
+++ b/tests/integration/components/render-content-test.js
@@ -3,7 +3,6 @@ import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 import { render } from '@ember/test-helpers';
 import hbs from 'htmlbars-inline-precompile';
-import { find } from 'ember-native-dom-helpers';
 
 module('Integration | Component | render-content', function(hooks) {
   setupRenderingTest(hooks);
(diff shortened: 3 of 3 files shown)


addon/templates/components/render-content.hbs
```diff
@@ -3,5 +3,9 @@
 {{else if contentIsComponentDefinition}}
   {{component content}}
 {{else}}
-  {{content}}
+  {{#if hasBlock}}
```
andrewpye: Any reason not to move this as an `else if` on the previous line?

vladucu: sure, why not

vladucu:
> I'm struggling to see what the use case for this is as it's already possible to just pass a component (or a hash of componentName and props) to render-content? 🤔

The main reason is to be able to provide a sane default, let's take my example from the description above: without this, how can I support `title` as a String or Component defaulted to a specific text and rendered as a specific component? (asking for real, maybe I missed something and it's totally possible)

andrewpye:
>how can I support `title` as a String or Component defaulted to a specific text and rendered as a specific component?

To confirm that I'm understanding correctly, you mean:

- you have a component that receives a `title` attribute which could be either a string, or some other component definition;
- if the title is a component definition, you want to render that component;
- if the title is a string, you want to render that string using some other component.

?

If that's the case, you could try using the [`is-component-definition`](https://github.com/smile-io/ember-polaris/blob/master/addon/helpers/is-component-definition.js) helper to do something like this (disclaimer: _I haven't actually tried this code_ 😅):

```hbs
{{render-content (if (is-component-definition title)
  title
  (component "polaris-subheading" text=title)
)}}
```

or alternatively in your component file you could define something like

```js
titleComponent: computed('title', function() {
  let title = this.get('title');
  if (typeof title === 'string') {
    return {
      componentName: 'polaris-subheading',
      props: {
        text: title,
      },
    };
  }

  return title;
}).readOnly(),
```

and then in the template just use `{{render-content titleComponent}}`.

I appreciate that both of these are slightly more long-winded, but on the other hand I think that it keeps the usage of `render-content` really explicit - I'm not super-keen on having `render-content` take a block which _maybe_ gets rendered, or maybe doesn't, depending on the type of the content passed in, since it requires any future maintainers to have the knowledge of when the block gets used or not, which isn't particularly intuitive just from seeing the usage.

As an alternative, what do you think about taking an optional named property allowing us to do something like this:

```hbs
{{render-content title
  fallbackComponent=(component "polaris-subheading" text=title)
}}
```

? I still think I prefer the more explicit approaches I suggested though.

vladucu:
You just proved my points with your suggestions 😄 

> I'm not super-keen on having render-content take a block which maybe gets rendered, or maybe doesn't, depending on the type of the content passed in, since it requires any future maintainers to have the knowledge of when the block gets used or not, which isn't particularly intuitive just from seeing the usage.

I feel like this is not really an argument...most of our components take a `text` param that's ignored if the component is used as a block.
I thought the whole purpose of this component was to allow rendering based on `content`, either a component or something else....I think with this still pretty much is the same idea: 
- if `content` is a component/component hash - renders it
- otherwise, if inline-mode, renders `content` directly. if block-mode, yields
 
The `fallbackComponent` suggestion is also rather limited, maybe I want to have more elements rendered when nothing is passed/string.

This being said, I can tweak the `is-component-definition` helper too instead of this (which doesn't supports component as hash)

PR #279, opened 2019-02-21
Ensure choice list updates properly when "selected" array is swapped out

andrewpye:
We found that swapping the array passed to `polaris-choice-list` as the `selected` property didn't trigger an update of which options were marked as selected. This should fix the issue... 🙏


andrewpye:
Putting this on hold because the change seems to have broken some existing tests 😬

PR #280, opened 2019-02-22
Allow passing a component as polaris-choice-list's helpText property

andrewpye:
This is supported by the React implementation, so we should probably support it too 😉


andrewpye:

PR #281, opened 2019-02-22
Handle non-numeric progress values in progress bar

andrewpye:
Because the React Polaris components use TypeScript, their `ProgressBar` component handles non-numeric values quite cleanly. Our implementation unfortunately did not, so in this PR I've updated its behaviour to match the Shopify version's:

- non-numeric `progress` values result in the bar being rendered without a `value` attribute, making it appear empty;
- Values of +/-infinity are now clamped to 100 and 0 internally respectively.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # ember-polaris Changelog
 
+### Unpublished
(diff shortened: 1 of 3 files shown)


andrewpye:

PR #282, opened 2019-02-25
Allow passing component for annotated layout section's description

andrewpye:
Until now, `polaris-layout/annotated-section`'s `annotation` supported being used in block form, with the block content being used in place of the `description` property. However, this block usage isn't accessible via `polaris-layout/annotated-section`'s public API; as a result I'm removing this behaviour in favour of accepting a component definition or `componentName`/`props` hash as the `description` to allow rendering arbitrary content directly via the public API.

### DIFF
--- a/addon/templates/components/polaris-layout/annotation.hbs
+++ b/addon/templates/components/polaris-layout/annotation.hbs
@@ -1,14 +1,14 @@
 {{#polaris-text-container}}
   {{polaris-heading text=title}}
 
-  {{#if (or hasBlock description)}}
+  {{#if description}}
     <div data-test-annotation-description class="Polaris-Layout__AnnotationDescription">
-      {{#if hasBlock}}
-        {{yield}}
-      {{else}}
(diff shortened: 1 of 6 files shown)


andrewpye:

addon/templates/components/polaris-layout/annotation.hbs
```diff
@@ -1,14 +1,14 @@
 {{#polaris-text-container}}
   {{polaris-heading text=title}}
 
-  {{#if (or hasBlock description)}}
+  {{#if description}}
     <div data-test-annotation-description class="Polaris-Layout__AnnotationDescription">
-      {{#if hasBlock}}
-        {{yield}}
-      {{else}}
+      {{#if hasStringDescription}}
```
vladucu: do we need this? wouldn't `render-content` handle this correctly?

andrewpye: We need to explicitly add the `p` wrapper if `description` is just a string

PR #283, opened 2019-02-28
Fix height bug on multiline text field

andrewpye:
Noticed that multiline text fields were rendering a couple of lines taller than expected. This was something we'd fixed internally at @smile-io but forgotten about, and recently got exposed in our apps during some development work 🤷‍♂️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # ember-polaris Changelog
 
+### Unreleased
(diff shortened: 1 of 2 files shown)


andrewpye:
>Weeeird 🤷‍♂️

@tomnez we had a second implementation of the resizer component in one of our internal addons from before you added `polaris-text-field`. That implementation had this fix. I just dropped that duplicate component since I knew we had this one here now, and that's when the issue showed up 😉

PR #284, opened 2019-03-01
Break infinite loop in sticky manager

andrewpye:
Noticed today while testing `polaris-sticky` in one of our apps that the `manageStickyItems` method in `sticky-manager` has an infinite loop, figured I should _probably_ fix it 🤦‍♂️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # ember-polaris Changelog
 
+### Unreleased
+- [#284](https://github.com/smile-io/ember-polaris/pull/284) [FIX] Fix infinite loop in `sticky-manager` service
+
 ### v4.0.1 (February 28, 2019)
 - [#283](https://github.com/smile-io/ember-polaris/pull/283) [FIX] Fix height bug on multiline text field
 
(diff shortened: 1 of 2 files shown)


andrewpye:
>It seems by default Polaris's throttle will have the _func invoked on the trailing edge of the timeout only if the throttled function is invoked more than once during the wait timeout_

Not quite, from what I can see - this is only if both `leading: true` and `trailing: true` are specified. The default behaviour seems to be `trailing`.

>Do you think using throttle with immediate false would do the same?

I'll try this, you might be onto something 👍 

>PS: seems Polaris uses 40 as the timeout instead of 50 for the throttles

Nowadays yes, but [not on v2.11.0](https://github.com/Shopify/polaris-react/blob/v2.11.0/src/components/AppProvider/StickyManager.ts#L67) which we currently mimic...

andrewpye:
@vladucu updated (and now the diff's best viewed with [whitespace diffing disabled](https://github.com/smile-io/ember-polaris/pull/284/files?utf8=%E2%9C%93&diff=split&w=1))

PR #286, opened 2019-03-20
Update avatar to Polaris v3.10.0

andrewpye:
Updates `polaris-avatar` to match Polaris v3.10.0. I haven't added the new `isServer` functionality since we don't need it at this point. Also dropped the old tests in favour of porting the new ones added in `polaris-react` since we initially built `polaris-avatar`.

### DIFF
--- a/addon/templates/components/polaris-avatar.hbs
+++ b/addon/templates/components/polaris-avatar.hbs
@@ -22,7 +22,7 @@
     src={{finalSource}}
     alt=""
     role="presentation"
-    {{action "handleLoad" on="load"}}
-    {{action "handleError" on="error"}}
+    onload={{action "handleLoad"}}
+    onerror={{action "handleError"}}
   >
 {{/if}}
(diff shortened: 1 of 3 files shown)


andrewpye:
[Kick-off ember-polaris upgrade to 3.x](https://trello.com/c/orVDHTEo/320-kick-off-ember-polaris-upgrade-to-3x)

addon/templates/components/polaris-avatar.hbs
```diff
@@ -15,12 +15,15 @@
...
-{{#if finalSource}}
+{{#if shouldShowImage}}
   {{!-- TODO: replace this with polaris-image when implemented. --}}
   <img
+    {{! template-lint-disable no-invalid-interactive }}
     class="Polaris-Avatar__Image"
     src={{finalSource}}
     alt=""
     role="presentation"
+    onload={{action "handleLoad"}}
```
vladucu: just double-checking these are the right events here in Ember world

andrewpye: Well, I initially went with `{{action "handleLoad" on="load"}}` element modifiers and tried tinkering with that, but apparently that approach doesn't work here so this is what I landed on. The updated tests cover this functionality and with the exception of a timing issue in the tests for Ember 2.16/2.18, they pass without problem.

PR #288, opened 2019-03-20
Update banner to Polaris v3.10.0

andrewpye:
Tweaks `polaris-banner` to add the new `div` around the primary action, and sets the primary action button size to `slim` when `withinContentContainer` is truthy. Also added a `TODO` to update the icons when we figure out how to progress with the `@shopify/polaris-icons` package 🙏


andrewpye:
[Kick-off ember-polaris upgrade to 3.x](https://trello.com/c/orVDHTEo/320-kick-off-ember-polaris-upgrade-to-3x)

PR #291, opened 2019-03-22
Update button to Polaris v3.10.0

andrewpye:
Adds `monochrome`, `download` and `isPressed` properties to `polaris-button`. Also, to make future updates easier, I've rewritten `polaris-button` from scratch as a more-or-less direct port of the React implementation 🤷‍♂️ This will need careful and thorough testing in our apps to ensure the rewrite hasn't broken anything, although all our old `polaris-button` tests still passed before I replaced them with the React tests.


andrewpye:

addon/templates/components/polaris-button.hbs
```diff
@@ -1,63 +1,80 @@
...
-        {{#if (is-component-definition icon)}}
-          {{render-content icon}}
+  {{#if url}}
+    {{#if isDisabled}}
+      {{!--
+        Render an `<a>` so toggling disabled/enabled state changes only the
+        `href` attribute instead of replacing the whole element.
+      --}}
+      <a id={{id}} class={{classes}} aria-label={{accessibilityLabel}} data-test-id={{dataTestId}}>
+        {{#if hasBlock}}
```
tomnez: Does

```handlebars
{{#component contentComponent}}
  {{yield}}
{{/component contentComponent}}
```

do the same thing as

```handlebars
{{component contentComponent}}
```

if there isn't a block to yield? If so maybe we don't need the `hasBlock` check.

andrewpye: I'd imagine most of the time, yes, but if `contentComponent` has different behaviour depending on whether it's in block or inline mode then we need to render it in either block or inline mode as appropriate. This is the case with quite a number of the components in `ember-polaris`, which yield if they're used in block mode, or render their `text` attribute otherwise.

I made a [twiddle](https://ember-twiddle.com/3fa614c4930f5c22faccecad45f9c112?openFiles=templates.application.hbs%2C) to show what I mean: both instance of `outer-component` cause `inner-component` to be rendered in inline mode - even if `outer-component` is rendered in block mode. It's a nuisance as your suggested simplification would be very nice, but unfortunately has breaking behaviour (which IIRC we've hit in the early days of `ember-polaris`) 😞

andrewpye:
Will wait for @vladucu's 👀 on this before merging cos this is a big and potentially breaking refactor 😅

addon/templates/components/polaris-button.hbs
```diff
@@ -1,63 +1,80 @@
...
-    {{/if}}
-
-    {{#if icon}}
-      <span class="Polaris-Button__Icon">
-        {{#if (is-component-definition icon)}}
-          {{render-content icon}}
+  {{#if url}}
+    {{#if isDisabled}}
+      {{!--
+        Render an `<a>` so toggling disabled/enabled state changes only the
```
vladucu: is this really a problem in Ember...or is it more to keep implementation as close as possible?

tests/integration/components/polaris-button-test.js
```diff
@@ -1,639 +1,333 @@
...
+      test('renders a link when present', async function(assert) {
+        await render(hbs`{{polaris-button url=mockUrl}}`);
+        assert.dom('a[data-polaris-unstyled]').exists();
+      });
+
+      test('gets passed into the link', async function(assert) {
+        await render(hbs`{{polaris-button url=mockUrl}}`);
+        assert
+          .dom('a[data-polaris-unstyled]')
+          .hasAttribute('href', this.get('mockUrl'));
```
vladucu: no biggie (or need to update), but we should really start to use ES6 getters

tests/integration/components/polaris-button-test.js
```diff
@@ -1,639 +1,333 @@
...
-  );
-});
+    module('children', function() {
+      test('renders the given children into the button', async function(assert) {
+        this.set('label', 'Click me!');
+        await render(hbs`{{#polaris-button}}{{label}}{{/polaris-button}}`);
+        assert.dom('button').hasText(this.get('label'));
+      });
+
+      test('renders the given children into the link', async function(assert) {
```
vladucu: `children` is really react-world.... 😜 
`block-form` ?

addon/templates/components/polaris-button.hbs
andrewpye: The latter - I don't see any real disadvantage to having this implemented and it makes sure we're matching the React implementation closely, which can't hurt 🤷‍♂️

tests/integration/components/polaris-button-test.js
andrewpye: Do ES6 getters work with older versions of Ember (e.g. 2.18, which we still support)? Or do they cause any potential issues with apps that use older versions of Ember?

tests/integration/components/polaris-button-test.js
andrewpye: 🤷‍♂️I think we all know what this means, and matching the React code makes the upgrade process much clearer (as I'm finding out this time around 😬)

PR #292, opened 2019-03-22
Update card to Polaris v3.10.0

andrewpye:
Tweaks logic behind when card header renders and prevents potential glimmer errors when card has dynamic footer actions.

EDIT: also adds support for rendering components as card section titles (missed this file in the diff when I made the initial changes 🤦‍♂️)

### DIFF
--- a/addon/templates/components/polaris-card.hbs
+++ b/addon/templates/components/polaris-card.hbs
@@ -1,4 +1,5 @@
 {{#if (or title headerActions)}}
+  {{!-- TODO: check for children here as per the React implementation. --}}
   {{#if headerActions}}
     <div class="Polaris-Card__Header">
       {{#polaris-stack alignment="baseline" as |stack|}}
--- a/addon/templates/components/polaris-card.hbs
+++ b/addon/templates/components/polaris-card.hbs
@@ -1,4 +1,4 @@
-{{#if title}}
+{{#if (or title headerActions)}}
   {{#if headerActions}}
     <div class="Polaris-Card__Header">
       {{#polaris-stack alignment="baseline" as |stack|}}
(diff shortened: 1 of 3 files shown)


andrewpye:

addon/templates/components/polaris-card.hbs
```diff
@@ -1,4 +1,4 @@
-{{#if title}}
+{{#if (or title headerActions)}}
```
tomnez: Any reason this is wrapped in two conditions instead of doing

```handlebars
{{#if headerActions}}
  <!-- do the things in the first block -->
{{else if title}}
  <!-- do the things in the second block -->
{{/if}}
```

?

andrewpye: Yes.

😝 

To expand, this came from mimicking the React implementation, which has a [`title || actions` condition](https://github.com/Shopify/polaris-react/blob/master/src/components/Card/Card.tsx#L53) in the main `Card` component which determines whether the `Header` subcomponent gets rendered 🤷‍♂️ I think this way is clearer when it comes to future maintenance but am open to changing things if you think there's value.

addon/templates/components/polaris-card.hbs
```diff
@@ -1,4 +1,4 @@
...
   {{#if headerActions}}
```
vladucu: nit: hmm, I guess we're not matching Polaris exactly here, which has this in a [`Header`](https://github.com/Shopify/polaris-react/blob/master/src/components/Card/components/Header/Header.tsx#L29) subcomponent and also checks if it has children

no need to fix this now....can we add a TODO here for that? I guess this entire component is a good candidate for refactoring soon too

andrewpye: Yeah exactly, would def want to rebuild this component as and when time allows, it's one from the early days so would be great to bring it in line with the React implementation 🙏

PR #293, opened 2019-03-22
Update checkbox to Polaris v3.10.0

andrewpye:
Updates `polaris-checkbox` error handling and adds focus fix as per the React implementation. Also updates the tests to match those found in the React implementation.

### DIFF
--- a/addon/components/polaris-checkbox.js
+++ b/addon/components/polaris-checkbox.js
@@ -34,6 +34,7 @@ export default Component.extend({
    * @property labelComponent
    * @type {String | Component}
    * @default null
+   * @deprecated
    * @public
    */
   labelComponent: null,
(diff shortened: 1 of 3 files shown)


andrewpye:

addon/components/polaris-checkbox.js
```diff
@@ -34,6 +33,7 @@ export default Component.extend({
    * @property labelComponent
    * @type {String | Component}
    * @default null
+   * @deprecated
    * @public
    */
   labelComponent: null,
```
vladucu: thoughts on actually making this a [deprecated CP](https://api.emberjs.com/ember/3.8/functions/@ember%2Fobject%2Fcomputed/deprecatingAlias) instead of just adding jsdocs?

andrewpye: Not sure this will add anything given that if an app passes in a `labelComponent` attribute it'll overwrite the CP, no? This has been deprecated for a while now, and we've had a more "solid" [deprecation warning in `didReceiveAttrs`](https://github.com/smile-io/ember-polaris/blob/00334db3e3cce3a724665a9667dbc90cebb612f4/addon/components/polaris-checkbox.js#L211) since then 🤷‍♂️

vladucu: ahh, good than

PR #294, opened 2019-03-22
Update choice property descriptions

andrewpye:
BIG CHANGES ALERT! jk, I diffed the React version against the wrong revision and thought there were some real changes to make, then by the time I realised I'd already updated these and thought we may as well keep 'em 🤷‍♂️


andrewpye:

PR #297, opened 2019-03-26
Update data table to Polaris v3.10.0

andrewpye:
Adds the new `data-polaris-header-cell` attribute to `polaris-data-table` cells with `header=true`.


andrewpye:

vladucu:
the diff gives me a bunch of changes....are those just shifting things around? nothing really changed in functionality?

andrewpye:
>the diff gives me a bunch of changes....are those just shifting things around? nothing really changed in functionality?

From what I could tell, the vast majority of the diff comes from a) moving things around b) adding i18n stuff, which works out with the same result as what we've already got here.

andrewpye:
@vladucu but if you think there's something I've missed, let me know! 🙏

andrewpye:
@vladucu yep got the right one. Made a couple of tweaks, wondering if there's anything else you think I've missed?

andrewpye:
@vladucu _bump_ 😅

vladucu:
ops, sorry
on it in 10 mins

addon/utils/data-table.js
```diff
@@ -0,0 +1,50 @@
+export function measureColumn(tableData) {
```
vladucu: Some slight difference to the original here
Screen Shot 2019-04-04 at 9 01 27 PM

andrewpye:
> Is there anything we need to update related to the changes below
> Screen Shot 2019-04-04 at 8 56 20 PM
> ?

The `sortDirection = defaultSortDirection` change doesn't seem to be important since they reassign `sortDirection` before it gets used 🤷‍♂️Adding it anyway though in case a future change in this area needs it and we miss it.

`sorted` doesn't seem to be used at all, so I'm just going to remove it #_yolo_ 😅

vladucu:
> The sortDirection = defaultSortDirection change doesn't seem to be important since they reassign sortDirection before it gets used 🤷‍♂️Adding it anyway though in case a future change in this area needs it and we miss it.
sorted doesn't seem to be used at all, so I'm just going to remove it #yolo 😅

figured it's something along those lines 👍

PR #298, opened 2019-03-26
Update date picker to Polaris v3.10.0

andrewpye:
Tweaks the logic on `polaris-date-picker`'s `inHoveringRange` to match the React implementation.


andrewpye:

andrewpye:
>a bit hard to follow, since we departed from the original implementation a bit

Agreed, this is one of the reasons I've been so keen to port the React implementations as closely as possible even when it's not necessarily needed...

PR #299, opened 2019-03-26
Update drop zone to Polaris v3.10.0

andrewpye:
Adds the React implementation's [reset inputHTMLElement to allow duplicate files](https://github.com/Shopify/polaris-react/commit/a81f09471f91be9a8f42d87dd6a572db9984b0f4) change.

### DIFF
--- a/addon/components/polaris-drop-zone.js
+++ b/addon/components/polaris-drop-zone.js
@@ -356,15 +356,19 @@ export default Component.extend(ContextBoundEventListenersMixin, {
       onDrop,
       onDropAccepted,
       onDropRejected,
+      allowMultiple,
       state,
     } = this.getProperties(
       'disabled',
       'onDrop',
       'onDropAccepted',
       'onDropRejected',
+      'allowMultiple',
       'state'
     );
-    if (disabled) {
+    let numFiles = state.get('numFiles');
+
+    if (disabled || (!allowMultiple && numFiles > 0)) {
       return;
     }
     let fileList = getDataTransferFiles(event);
@@ -377,8 +381,8 @@ export default Component.extend(ContextBoundEventListenersMixin, {
     state.setProperties({
       dragging: false,
       error: rejectedFiles.length > 0,
+      numFiles: state.get('numFiles') + acceptedFiles.length,
     });
-    state.incrementProperty('numFiles', acceptedFiles.length);
 
     onDrop(files, acceptedFiles, rejectedFiles);
 
@@ -403,6 +407,7 @@ export default Component.extend(ContextBoundEventListenersMixin, {
       'allowMultiple',
       'state'
     );
+
     if (disabled || (!allowMultiple && state.get('numFiles') > 0)) {
       return;
     }
@@ -436,13 +441,13 @@ export default Component.extend(ContextBoundEventListenersMixin, {
       'onDragOver',
       'allowMultiple'
     );
+
     if (disabled || (!allowMultiple && state.get('numFiles') > 0)) {
       return;
     }
 
     onDragOver();
 
-    // Browsers are somewhat inconsistent about needing these, but they don't hurt to add.
     return false;
   },
 
@@ -456,6 +461,7 @@ export default Component.extend(ContextBoundEventListenersMixin, {
       'allowMultiple',
       'dropNode'
     );
+
     if (disabled || (!allowMultiple && state.get('numFiles') > 0)) {
       return;
     }
--- a/addon/components/polaris-drop-zone.js
+++ b/addon/components/polaris-drop-zone.js
@@ -386,6 +386,8 @@ export default Component.extend(ContextBoundEventListenersMixin, {
     if (rejectedFiles.length) {
       onDropRejected(rejectedFiles);
     }
+
+    event.target.value = '';
   },
 
   handleDragEnter(event) {
--- a/addon/components/polaris-drop-zone.js
+++ b/addon/components/polaris-drop-zone.js
@@ -356,11 +356,13 @@ export default Component.extend(ContextBoundEventListenersMixin, {
       onDrop,
       onDropAccepted,
       onDropRejected,
+      state,
     } = this.getProperties(
       'disabled',
       'onDrop',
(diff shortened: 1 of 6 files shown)


andrewpye:

andrewpye:

vladucu:
There are a bunch of other changes here from what I can see, or I'm wrong?

andrewpye:
>There are a bunch of other changes here from what I can see, or I'm wrong?

Wondering what those changes are in case I missed anything - just checked the diff again and the majority of the changes are import restructuring and updates to use the icon code changes which we don't have at this point (and _can't_ have at this point, see https://github.com/Shopify/polaris-react/issues/385 for details).

vladucu:
> Wondering what those changes are in case I missed anything

https://github.com/Shopify/polaris-react/compare/v2.11.0...v3.10.0#diff-7ee058419e736ffe4c643519496b920fR426

andrewpye:
lol... the diff is too big to show me what you're pointing out 😅 Were there any lines in particular that worried you? I'll accept a screen grab since github can't handle such large diffs 😉

vladucu:
screen-shot
Screen Shot 2019-03-27 at 12 53 35 PM

andrewpye:
@vladucu updated with the changes I missed earlier - thanks for spotting those! 🙌 

There are a handful of new tests that were added for `overlayText` and `errorOverlayText` on different screen sizes - I'll add them in the morning 👍

andrewpye:
@tomnez yeah I originally left those file name updates as part of the bigger icon update that we'll need to do in the future, but realised there wasn't really any reason not to update them now so that's done.

I've also added the tests I referred to yesterday, so this should be complete now 🙏

addon/components/polaris-drop-zone.js
```diff
@@ -356,11 +356,13 @@ export default Component.extend(ContextBoundEventListenersMixin, {
...
+      state,
     } = this.getProperties(
       'disabled',
       'onDrop',
       'onDropAccepted',
-      'onDropRejected'
+      'onDropRejected',
+      'state'
     );
     if (disabled) {
```
vladucu: should this be something like
`if (disabled || (!allowMultiple && numFiles > 0)) {`

addon/components/polaris-drop-zone.js
```diff
@@ -372,10 +374,11 @@ export default Component.extend(ContextBoundEventListenersMixin, {
 
     this.dragTargets = [];
 
-    this.get('state').setProperties({
+    state.setProperties({
       dragging: false,
       error: rejectedFiles.length > 0,
     });
+    state.incrementProperty('numFiles', acceptedFiles.length);
```
vladucu: nit: should we just move in the `setProperties` above instead of using `incrementProperty`?

addon/templates/components/polaris-drop-zone/file-upload.hbs
```diff
@@ -13,11 +13,11 @@
 {{else if (eq size "large")}}
   {{#polaris-stack vertical=true spacing="tight"}}
     {{#if (eq type "file")}}
-      {{svg-jar assetFileUpload class=imageClasses}}
+      {{svg-jar fileUpload class=imageClasses}}
```
vladucu: this might be my bad when initially implemented....but instead of an `svg` elements, we should be using `img`

vladucu: don't think there was a specific reason for doing this

tomnez:
@vladucu maybe you're looking at an older diff or something? I see the changes in `addon/components/polaris-drop-zone/file-upload.js` when I visit the diff tab.

vladucu:
hmm, @tomnez I only see some renaming
Screen Shot 2019-03-28 at 8 40 19 PM

tomnez:
@vladucu yeah that was the change I requested. What else were you thinking was there?

Although it looks like the rename of `iconDragDrop` to `DragDropMajorMonotone` was missed 😕

andrewpye:
>Although it looks like the rename of `iconDragDrop` to `DragDropMajorMonotone` was missed 😕

@tomnez this was deliberate - since Polaris switched to using `@shopify/polaris-icons`, all of their icon SVGs have changed name. We can't currently take those changes (discussion with Shopify is [here](https://github.com/Shopify/polaris-react/issues/385)) so we need to keep using the same old icon names for the time being, until we can start using the new icon formats.

vladucu:
> @tomnez this was deliberate - since Polaris switched to using @shopify/polaris-icons, all of their icon SVGs have changed name. We can't currently take those changes (discussion with Shopify is here) so we need to keep using the same old icon names for the time being, until we can start using the new icon formats.

ohh, so only the icon names changed here....but the actual icon didnt?

andrewpye:
>ohh, so only the icon names changed here....but the actual icon didnt?

TBH I haven't checked the new icon, but from what I've seen they've just changed the naming to be more explicit, and capitalised it to reflect the fact that the icons are now React components instead of SVG files. The way I see it, we need to make a fairly hefty overhaul to `polaris-icon` which we can't do yet so I'm thinking we leave icons as they are for now and revisit icon usages at the time of that update, thoughts?

addon/components/polaris-drop-zone.js
andrewpye: Yes you're right... GitHub somehow had those changes collapsed so I missed them 🤦‍♂️

vladucu:
> TBH I haven't checked the new icon, but from what I've seen they've just changed the naming to be more explicit, and capitalised it to reflect the fact that the icons are now React components instead of SVG files. The way I see it, we need to make a fairly hefty overhaul to polaris-icon which we can't do yet so I'm thinking we leave icons as they are for now and revisit icon usages at the time of that update, thoughts?

works for me...maybe add a TODO so we don't forget?

andrewpye:
>works for me...maybe add a TODO so we don't forget?

Should be pretty clear anyway, we'll need to update _all_ `polaris-icon` usages so a TODO seems unnecessary 🤷‍♂️

PR #300, opened 2019-03-26
Update empty search result to Polaris v3.10.0

andrewpye:
Tweaks the description on one of the `polaris-empty-search-result` tests, no other changes needed to match the React implementation 🤷‍♂️


andrewpye:

andrewpye:
>I think we should stop trying to match 💯 Polaris tests

I already had a branch open for updating this component after thinking there were more changes we needed to make, and figured I may as well use it 😝 

On a slightly more serious note, I do think it's worth having the Polaris tests matched fairly closely, since a) it highlights what behaviours Shopify consider to be important, and b) makes it easier to update our own tests as and when there's new functionality or a bug fix introduced to a component.

PR #304, opened 2019-03-27
Update link to Polaris v3.10.0

andrewpye:
Adds support for `monochrome` `polaris-link`s and ensures the correct `type` is used when link buttons are rendered.

### DIFF
--- a/tests/integration/components/polaris-link-test.js
+++ b/tests/integration/components/polaris-link-test.js
@@ -92,6 +92,25 @@ test('it renders the correct HTML with external attribute', function(assert) {
   );
 });
 
+test('it renders the correct HTML with monochrome attribute', function(assert) {
+  this.render(hbs`
+    {{polaris-link
+      url="http://www.somewhere.com/"
+      monochrome=true
+      text="Testing monochrome attribute"
+    }}
+  `);
+
+  const links = findAll(linkSelector);
+  assert.equal(links.length, 1, 'renders one link');
+
+  const link = links[0];
+  assert.ok(
+    link.classList.contains('Polaris-Link--monochrome'),
+    'sets the monochrome class on the link'
+  );
+});
+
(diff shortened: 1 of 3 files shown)


andrewpye:

tests/integration/components/polaris-link-test.js
```diff
@@ -92,6 +92,25 @@ test('it renders the correct HTML with external attribute', function(assert) {
   );
 });
 
+test('it renders the correct HTML with monochrome attribute', function(assert) {
+  this.render(hbs`
```
vladucu: would it be too much to ask to convert the test to new syntax? 😬 
we can leave for later too, but would be nice to do it while here

andrewpye: I'd prefer to get the update done in one go as quickly as possible so we can start using it, and update tests _en masse_ in the future - they're separate changes in my mind.

vladucu: as you think it's best (anything needed for the rush?)

from my experience, that en masse = never lmao

andrewpye: >anything needed for the rush?

Updating the Polaris version takes a long time and we'll almost certainly have other projects coming up which will divert resources away from the update, and potentially add new parts of our apps that need updating. Test syntax updates can be done either in bulk or on an individual file basis moving forward so they're less "intrusive" changes, and don't really add any value as they're more of a "sideways" move. Getting the update out and in use seems like the more pressing change at this point 🤷‍♂️

vladucu: good with me

though, again, that's the reason we kicked of this update.... _nothing should be pressing_

andrewpye: Maybe my wording oversold the urgency of this 😂 But I still feel like a better approach is to get the code changes _required_ for the Polaris update out of the way first in one go, then make further updates as and when we have the cycles to spare 🙏

PR #306, opened 2019-03-29
Update resource list to Polaris v3.10.0

andrewpye:
Adds support for `alternateTool` to `polaris-resource-list`, improves handling of loading state, improves item link handling on click events, and updates the resource list subcomponents with the tweaks from Polaris v3.10.0.

### DIFF
--- a/addon/components/polaris-resource-list/item.js
+++ b/addon/components/polaris-resource-list/item.js
@@ -16,6 +16,7 @@ export default Component.extend(context.ConsumerMixin, {
     'persistActions:Polaris-ResourceList-Item--persistActions',
     'focusedInner:Polaris-ResourceList-Item--focusedInner',
   ],
+  attributeBindings: ['url:data-href'],
 
   layout,
 
@@ -221,6 +222,7 @@ export default Component.extend(context.ConsumerMixin, {
       'selectMode',
       'element'
     );
+    let { ctrlKey, metaKey } = event;
     let anchor = element && element.querySelector('a');
 
     if (selectMode) {
@@ -234,6 +236,11 @@ export default Component.extend(context.ConsumerMixin, {
 
     if (onClick) {
       onClick(itemId);
+    }
+
+    if (url && (ctrlKey || metaKey)) {
+      window.open(url, '_blank');
+      return;
     }
 
     if (url && anchor) {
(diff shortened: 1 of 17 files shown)


andrewpye:

vladucu:
Screen Shot 2019-04-02 at 10 17 30 PM

🙄

addon/components/polaris-resource-list/item.js
```diff
@@ -221,6 +222,7 @@ export default Component.extend(context.ConsumerMixin, {
       'selectMode',
       'element'
     );
+    let { ctrlKey, metaKey } = event;
```
vladucu: should this be `event.nativeEvent`?

addon/templates/components/polaris-resource-list/item.hbs
```diff
@@ -69,7 +73,7 @@
     </div>
   {{/if}}
 
-  {{#if shortcutActions}}
+  {{#if (and shortcutActions (not loading))}}
     {{#wrapper-element
```
vladucu: why do we use this `wrapper-element` here?

addon/components/polaris-resource-list/item.js
andrewpye: No, React wraps all event objects in a [`SyntheticEvent`](https://reactjs.org/docs/events.html) class, whose `nativeEvent` property is the underlying browser event. We don't seem to have to worry about such things in Ember-land 🙌

addon/templates/components/polaris-resource-list/item.hbs
andrewpye: I think to handle the `click` and gain access to the event so that we can [stop it propagating](https://github.com/smile-io/ember-polaris/blob/3861a68eef235a99a42346bafdb320dc6eab44aa/addon/templates/components/polaris-resource-list/item.hbs#L80) I think, this was from a while ago 😅

addon/components/polaris-resource-list/item.js
vladucu: sweet 🙌

addon/templates/components/polaris-resource-list/item.hbs
vladucu: kk

PR #309, opened 2019-04-02
Update text style to Polaris v3.10.0

andrewpye:
Updates the elements used so that `span` is now used when `variation` is `'strong'`, and DRYs up the template slightly by using `wrapper-element`.


andrewpye:

PR #310, opened 2019-04-02
Update textfield to Polaris v3.10.0 

tomnez:
Adds a new `showCharacterCount` attribute as well as methods for holding down the buttons in the number spinners to slowly increment/decrement the current number value.

#### Character count with max length:
character-count-multiline-maxlen

#### Character count with no max length:
character-count-multiline

#### Press and hold spinner:
spinner-hold

### DIFF
--- a/addon/components/polaris-text-field.js
+++ b/addon/components/polaris-text-field.js
@@ -9,7 +9,6 @@ import { getCode } from 'ember-keyboard';
 import layout from '../templates/components/polaris-text-field';
 import { normalizeAutoCompleteProperty } from '../utils/normalize-auto-complete';
 import ContextBoundEventListenersMixin from 'ember-lifeline/mixins/dom';
-import ContextBoundTasksMixin from 'ember-lifeline/mixins/run';
 
 /**
  * Returns the length of decimal places in a number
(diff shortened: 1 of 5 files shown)


addon/components/polaris-text-field.js
```diff
@@ -504,21 +515,88 @@ export default Component.extend({
...
-    let { input, focused } = this.getProperties('input', 'focused');
+  normalizedValue: computed('value', function() {
+    let value = this.get('value');
+    return value !== null ? value : '';
+  }),
 
-    if (input && focused) {
-      // TODO this _seems_ to work in Polaris without the row below, but not for us (it does not apply focus class)
-      // look what's here....though this seems to do the job for now
-      this.set('focus', true);
```
tomnez: This was removed in favor of the `focus` computed property below now that the react implementation has updated their [code](https://github.com/Shopify/polaris-react/blob/bba36af2ad10ba9626bba168b52e2e9222ac4c60/src/components/TextField/TextField.tsx#L149-L157) to check the previous and current value of `focused` in order to determine whether or not to focus or blur the field.

We can't really do the same thing and call `blur` on `didUpdateAttrs` because there's a chance _other_ attributes have been updated while `focused` is set to `false`, and we wouldn't want to accidentally blur the field as the user is typing just because the initial value of `focused` was `false`.

addon/components/polaris-text-field.js
```diff
@@ -504,21 +515,88 @@ export default Component.extend({
     return inputType === 'number' && !disabled;
   }).readOnly(),
 
-  checkFocus() {
-    let { input, focused } = this.getProperties('input', 'focused');
+  normalizedValue: computed('value', function() {
+    let value = this.get('value');
+    return value !== null ? value : '';
```
andrewpye: Might be worth explicitly matching the React implementation and using `!=` here so that this catches `undefined` values etc.

addon/components/polaris-text-field.js
```diff
@@ -504,21 +515,88 @@ export default Component.extend({
...
+  }),
 
-    if (input && focused) {
-      // TODO this _seems_ to work in Polaris without the row below, but not for us (it does not apply focus class)
-      // look what's here....though this seems to do the job for now
-      this.set('focus', true);
-      input.focus();
+  characterCount: computed('normalizedValue.length', function() {
```
andrewpye: This should maybe be `readOnly`

addon/components/polaris-text-field.js
```diff
@@ -504,21 +515,88 @@ export default Component.extend({
     return inputType === 'number' && !disabled;
   }).readOnly(),
 
-  checkFocus() {
-    let { input, focused } = this.getProperties('input', 'focused');
+  normalizedValue: computed('value', function() {
+    let value = this.get('value');
+    return value !== null ? value : '';
+  }),
```
andrewpye: Can this be `readOnly`?

addon/components/polaris-text-field.js
```diff
@@ -504,21 +515,88 @@ export default Component.extend({
...
+  characterCountLabel: computed('maxLength', 'characterCount', function() {
+    let { maxLength, characterCount } = this.getProperties(
+      'maxLength',
+      'characterCount'
+    );
+
+    return maxLength
+      ? `${characterCount} characters of ${maxLength} used`
+      : `${characterCount} characters`;
+  }),
```
andrewpye: `readOnly`?

addon/components/polaris-text-field.js
```diff
@@ -504,21 +515,88 @@ export default Component.extend({
...
+  characterCountClassName: computed('multiline', function() {
+    let classNames = ['Polaris-TextField__CharacterCount'];
+
+    if (this.get('multiline')) {
+      classNames.push('Polaris-TextField__AlignFieldBottom');
     }
-  },
+
+    return classNames.join(' ');
+  }),
```
andrewpye: `readOnly`?

addon/components/polaris-text-field.js
andrewpye: >We can't really do the same thing and call `blur` on `didUpdateAttrs` because there's a chance _other_ attributes have been updated while `focused` is set to `false`, and we wouldn't want to accidentally blur the field as the user is typing just because the initial value of `focused` was `false`.

Not sure I follow this 😬 Can't we track `wasFocused` internally and call `blur`/`focus` as appropriate from `didUpdateAttrs`? If `focused` changes, then we should still be obeying it, and if you're worried about "accidentally blurring the field as the user is typing", that's on the host app to pass sensible property values 🤷‍♂️

addon/components/polaris-text-field.js
```diff
@@ -504,21 +515,88 @@ export default Component.extend({
...
+    let { maxLength, characterCount } = this.getProperties(
+      'maxLength',
+      'characterCount'
+    );
+
+    return !maxLength ? characterCount : `${characterCount}/${maxLength}`;
+  }),
+
+  focus: computed('focused', function() {
+    return this.get('focused') || false;
```
andrewpye: lol, this is weird logic but I appreciate that it's in the React implementation 🤷‍♂️ Anyhoo, how 'bout making this `readOnly` too?

addon/components/polaris-text-field.js
```diff
@@ -504,21 +515,88 @@ export default Component.extend({
...
+    let onChangeInterval = () => {
+      if (interval > minInterval) {
+        interval -= decrementBy;
+      }
+
+      onChange();
+
+      this.set(
+        'buttonPressTimer',
+        window.setTimeout(onChangeInterval, interval)
```
andrewpye: I wonder if we're better off using runloop-compatible timer functions here 🤔 This is probably fine in React-land but it's making me nervous in Ember-world 😅

vladucu: 👍

addon/components/polaris-text-field.js
tomnez: I guess we can make some bootleg version of `wasFocused`. The original buggyness I was seeing wasn't a host app being irresponsible, it was that you couldn't type without the field blur-ing unless you explicitly passed `focused=true`, which shouldn't happen. I'll see about making our own `wasFocused`.

andrewpye: Yeah, I don't see why this shouldn't work with `wasFocused` (we're doing similar things in some [other components](https://github.com/smile-io/ember-polaris/blob/update/polaris-v3.10.0/addon/components/polaris-resource-list.js#L800) already)

addon/components/polaris-text-field.js
vladucu: ^this

addon/templates/components/polaris-text-field.hbs
```diff
@@ -45,7 +45,7 @@
           minlength=minLength
           maxlength=maxLength
           spellcheck=spellCheck
-          value=value
+          value=normalizedValue
```
vladucu: we should also pass `step` here?

addon/components/polaris-text-field/spinner.js
```diff
@@ -25,4 +25,17 @@ export default Component.extend({
    * @default noop
    */
   onClick() {},
+
+  onMouseDown() {},
+
+  onMouseUp(/* onChange */) {},
```
vladucu: nit: `onMouseDown` gets invoked with `onChange`, not `onMouseUp` 😜

vladucu:
> Any idea why the native spinner is showing here?
image
It doesn't seem to appear on the React implementation... 😬 

@tomnez any luck with this ^ ?

tomnez:
@vladucu ah missed that too, I'll look into it.

vladucu:
thx!

tomnez:
I'm having wacky node issues with linking the host apps to this branch, we suspect this is just a dummy app issue since the React markup also displays the same problem when copied into the dummy app so we're going to merge and make a note to double check this on a real app.

PR #311, opened 2019-04-03
Update page to Polaris v3.10.0

andrewpye:
Mostly updates properties + their descriptions, component structure and CSS classes 🤷‍♂️


andrewpye:

andrewpye:
>a bit hard to follow

Yeah totally agree, going through the React diff for this was a bit of a pain too 🤕 

>Assuming tests pass and it's a widely used component we should be able to catch if anything is missed, etc

Agreed here too - I know I've done _something_ right here because the dummy app looked like a train wreck before making these updates 😅

PR #312, opened 2019-04-03
Use ember-lifeline for button press/release handling

vladucu:
@tomnez something I came up with that from what I see matches Polaris functionality


tomnez:
@vladucu I cherry-picked this into my current PR, thanks for the idea! 💪💪💪

vladucu:
🌸

PR #314, opened 2019-04-08
Remove event object from button mouse action handler invocations

andrewpye:
https://github.com/smile-io/ember-polaris/pull/291 inadvertently changed the interface for `polaris-button`, such that the mouse event object was passed to the `onClick`, `onFocus` and `onBlur` actions on invocation, which it never was before my refactor. This change removes the erroneous event object, fixing the signature for `onClick`, `onFocus` and `onBlur` :+1:


andrewpye:
@vladucu added 👍

PR #318, opened 2019-05-10
Add support for choice-list extra component aka renderChildren

vladucu:
### What's here

This adds support for an `extra` component to be used for `polaris-choice-list` choices aka `renderChildren` in react world

### DIFF
--- a/addon/components/polaris-choice-list.js
+++ b/addon/components/polaris-choice-list.js
@@ -53,7 +53,7 @@ export default Component.extend({
    *  - label
    *  - disabled
    *  - helpText
-   *  - extra (Polaris's renderChildren equivalent)
+   *  - childComponent (Polaris's renderChildren equivalent)
    *
    * @property choices
    * @public
@@ -113,13 +113,13 @@ export default Component.extend({
   titleHidden: false,
 
   /**
-   * Component to render children with a choice.
+   * Component to render a child component for a choice.
    *
    * @type {Component}
-   * @property extra
+   * @property childComponent
    * @public
    */
-  extra: null,
+  childComponent: null,
 
   /**
    * Callback when the selected choices change
--- a/addon/components/polaris-choice-list.js
+++ b/addon/components/polaris-choice-list.js
@@ -3,7 +3,7 @@ import { get, computed } from '@ember/object';
 import { guidFor } from '@ember/object/internals';
 import { isEmpty } from '@ember/utils';
 import ObjectProxy from '@ember/object/proxy';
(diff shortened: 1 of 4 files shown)


addon/components/polaris-choice-list.js
```diff
@@ -112,6 +112,15 @@ export default Component.extend({
...
   titleHidden: false,
 
+  /**
+   * Component to render children with a choice.
+   *
+   * @type {Component}
+   * @property extra
+   * @public
+   */
+  extra: null,
```
andrewpye: Is this used? It doesn't seem to match the React implementation either, can we remove it?

addon/templates/components/polaris-choice-list.hbs
```diff
@@ -17,6 +17,12 @@
         checked=choice.isSelected
         onChange=(action "updateSelectedChoices" choice)
       }}
+
+      {{#if choice.extra}}
+        <div class="Polaris-ChoiceList__ChoiceChildren">
+          {{component choice.extra choice=choice}}
```
andrewpye: Can we call this something more meaningful and in line with the React implementation than `extra`, maybe `childComponent`? Also, to be in keeping with the React version, this should only pass `choice.isSelected` into the child component.

addon/components/polaris-choice-list.js
vladucu: it will be...that's why I'm adding it

addon/templates/components/polaris-choice-list.hbs
vladucu: if you think `childComponent` is more inline with react...sure....don't really care about naming here, since either way it won't match react version

vladucu:
@andrewpye updated

>  speaking of, would you mind updating the docs with an example of this in action please

added, though I think these docs don't have much value and pretty cumbersome to maintain.....and the tests should provide enough examples really

addon/components/polaris-choice-list.js
andrewpye: Why is it needed? And since it's not used yet, can we add it when it's needed - maybe in `@smile-io/ember-smile-polaris` since it's an extension to the original Polaris spec? 🤔

addon/templates/components/polaris-choice-list.hbs
andrewpye: It's certainly clearer as to its meaning that `extra` so yes - just because it won't ever match the React version exactly doesn't mean it shouldn't be clearly named 😜

PR #320, opened 2019-05-29
Ember version upgrade

tomnez:
Updates Ember to 3.10 and fixes deprecations introduced between 3.8 and 3.10.


addon/components/polaris-banner.js
```diff
@@ -183,7 +183,9 @@ export default Component.extend({
   didRender() {
     this._super(...arguments);
 
-    let hasContentWrapper = this.$().find('div.Polaris-Banner__Content').length;
+    let hasContentWrapper = this.element.querySelectorAll(
```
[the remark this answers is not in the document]
andrewpye: Can this be `querySelector` instead of `querySelectorAll`?

addon/components/polaris-popover/content.js
```diff
@@ -82,29 +81,40 @@ export default Component.extend({
       return;
     }
 
-    const trigger = Ember$(
+    const trigger = document.querySelector(
```
andrewpye: This will blow up if ever anybody tries using this addon in fastboot, where `document` is unavailable. Having said that, we probably have a bunch of stuff which is fastboot-unfriendly (anything using the `didInsertElement ` hook, for example, or JQuery (which maybe you've got rid of completely in this PR? 🙏)) so maybe let's not worry about it at this point and update for fastboot compatibility if we ever need it in the future?

addon/components/polaris-popover/content.js
```diff
@@ -82,29 +81,40 @@ export default Component.extend({
...
 
     const triggerRect = getRectForNode(trigger);
-    const left =
-      triggerRect.width / 2 + (triggerRect.left - getRectForNode(content).left);
-    Ember$('div.Polaris-Popover__Tip', content).css({ left });
+    const left = `${triggerRect.width / 2 +
+      (triggerRect.left - getRectForNode(content).left)}px`;
+    const toolTip = document.querySelector('div.Polaris-Popover__Tip');
+
+    if (toolTip) toolTip.style.left = left;
```
andrewpye: Can we avoid single-line, un-curlied `if`-statements please? 🙏

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -350,6 +338,18 @@ export default Component.extend(
       }
     ),
 
+    get moreActionsNode() {
```
andrewpye: Does this syntax work with `EmberObject` class declarations? 😮 If it does, how do you feel about updating the other CPs in this file so we have consistency - or is that not something we can do since these are volatile and the others aren't?

addon/utils/wrap-children.js
```diff
@@ -0,0 +1,34 @@
...
+    childNode = children[i];
+
+    if (shouldWrapFn(childNode)) {
+      clonedWrapper = wrapperNode.cloneNode();
+      childNode.parentNode.insertBefore(clonedWrapper, childNode);
+      clonedWrapper.appendChild(childNode);
+    }
+  }
+
+  wrapperNode.remove();
```
andrewpye: Is this needed? We shouldn't be inserting the wrapper element into the DOM at all, right? 🤔

addon/components/polaris-resource-list/bulk-actions.js
sivakumar-kailasam: Chiming in 🙂

This [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) is a js language functionality, so will work with any pojo.

[The ember deprecation for this](https://deprecations.emberjs.com/v3.x/#toc_computed-property-volatile) explains why this is a good idea for CPs marked volatile(TLDR, we invented volatile instead of recommending getters in the first place😅). In other cases we'll lose the advantage of accessing cached values if they were plain getters instead of CPs.

andrewpye: Thanks for the explanation @sivakumar-kailasam! 🙇

addon/components/polaris-button-group.js
```diff
@@ -66,11 +67,13 @@ export default Component.extend({
...
-      .not('div.Polaris-ButtonGroup__Item')
-      .wrap(
-        '<div data-test-button-group-item class="Polaris-ButtonGroup__Item"></div>'
-      );
+    let wrapper = document.createElement('div');
+
+    wrapper.classList.add('Polaris-ButtonGroup__Item');
+    wrapper.setAttribute('data-test-button-group-item', true);
+
+    wrapChildren(this.element.children, wrapper, function(el) {
```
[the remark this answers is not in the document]
tomnez: I'm good with this, when I refactor you might see a bit of a name change in some spots since some of the usages aren't only specific to `class` names, we also check for things like `role` in the form layout groups. But either way 👍

.eslintrc.js
```diff
@@ -2,7 +2,49 @@
 'use strict';
 
 module.exports = {
-  plugins: ['smile-ember'],
-  extends: ['plugin:smile-ember/addon'],
+  root: true,
```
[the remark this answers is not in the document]
tomnez: Ah thank you, I wasn't sure if the additions were needed or just upgrade noise.

addon/components/polaris-banner.js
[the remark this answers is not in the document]
tomnez: Ah you're right. It's only here to check if it's present in the DOM so it can just return 1 node (in fact I'm not sure if this would ever return more than 1 item anyway)

addon/utils/wrap-children.js
tomnez: Oh good catch 🤦‍♂Yeah it's not even in the DOM.

addon/components/polaris-popover/content.js
tomnez: Sure 😝

addon/utils/wrap-children.js
```diff
@@ -0,0 +1,38 @@
...
+    node.parentNode.insertBefore(clonedWrapper, node);
+    clonedWrapper.appendChild(node);
+  });
+}
+
+/**
+ * Convert htmlCollection to an array in order
+ * to use array methods on the collection.
+ */
+export function collectionToArray(htmlCollection) {
```
tomnez: FYI @sivakumar-kailasam splatting an `htmlCollection` into an array wasn't working as expected, so that's what this is here for.

addon/utils/wrap-children.js
```diff
@@ -1,34 +1,38 @@
...
-    if (shouldWrapFn(childNode)) {
-      clonedWrapper = wrapperNode.cloneNode();
-      childNode.parentNode.insertBefore(clonedWrapper, childNode);
-      clonedWrapper.appendChild(childNode);
-    }
+/**
+ * Convert htmlCollection to an array in order
+ * to use array methods on the collection.
+ */
+export function collectionToArray(htmlCollection) {
```
sivakumar-kailasam: do we need this fn over `[...htmlCollection]`?

addon/utils/wrap-children.js
sivakumar-kailasam: interesting, what did you notice?

tomnez: Turns out I was mistaken. The original reason this wasn't working wasn't because of splatting this into an array, it was because I was calling `reject` on a native array. So yeah I'll update.

addon/components/polaris-button-group.js
```diff
@@ -1,6 +1,6 @@
 import Component from '@ember/component';
 import layout from '../templates/components/polaris-button-group';
-import wrapChildren from '../utils/wrap-children';
+import wrapChildren, { childrenWithoutClassName } from '../utils/wrap-children';
```
andrewpye: @tomnez what do you think about renaming this util file to something like `dom` and making `wrapChildren` a named export instead of the default? Asking because `childrenWithoutClassName` doesn't really seem to belong in a util file with such a specific name as `wrap-children` IMO 🤔

addon/components/polaris-button-group.js
```diff
@@ -67,13 +67,14 @@ export default Component.extend({
     this._super(...arguments);
 
     // Wrap each child element that isn't already a group item.
+    let nodesToWrap = childrenWithoutClassName(
+      this.element.children,
```
andrewpye: The name `childrenWithoutClassName` implies to me that the function will take a node, and return a list of its children filtered by class name - but here it looks like it's actually returning a filtered copy of the passed-in list of nodes and not accessing any children. If that's the case, can we update the name to remove the `children` part - maybe something like `rejectNodesByClassName`?

vladucu:
😍

addon/components/polaris-button-group.js
tomnez: works for me

addon/components/polaris-button-group.js
tomnez: also works for me 😛

tomnez:
@andrewpye updated

tests/dummy/app/templates/test.hbs
```diff
@@ -1 +1,11 @@
-{{outlet}}
+{{#polaris-button-group as |group|}}
```
sivakumar-kailasam: Looks like this was an unintentional change.

PR #322, opened 2019-05-30
Upgrade tests

sivakumar-kailasam:
Removed `ember-native-dom-helpers` which was supposed to be a predecessor for `@ember/test-helpers`. Also refactored a bunch of tests to use the verbose `qunit-dom`api.

Additionally bumped up the minimum supported ember version to ember-source@3.10. So we should publish a major release of this addon once this is merged.


sivakumar-kailasam:
I wanted to merge this into Tom's PR & then add that note.

vladucu:
🚀

PR #323, opened 2019-05-30
Upgrade 2/3 - enable glimmer template-only components

tomnez:
Part 2 of https://github.com/smile-io/ember-polaris/pull/320.

Seems that this is all that's needed (?), I generated new components and it's using the `glimmer-component` blueprint now and they work with the JS files removed. Let me know if I've missed something.


tomnez:
@andrewpye, @sivakumar-kailasam was working on a separate PR to fix the failing tests so I've left those alone. Also Siva - do you know the answer to Andy's question? I feel like host apps should be fine, right?

sivakumar-kailasam:
I should've mentioned this earlier, but `@ember/optional/features` works slightly differently for addons. Checkout https://github.com/emberjs/ember-optional-features#at-build-time-from-an-addon

.ember-cli.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+process.env.EMBER_VERSION = "OCTANE";
```
sivakumar-kailasam: we aren't moving to octane yet 🙂

tomnez: Ah ok.. I hadn't seen that note about using glimmer components in an addon until just now. I had tried `ember feature:enable template-only-glimmer-components` after the upgrade and it added the feature to the `optional-features.json` but still wasn't generating glimmer components so I thought we needed this 😝I'll update

tomnez:
Closing because I don't think we need to do anything to have just `template-only` components on the upgrade branch.

andrewpye:
@tomnez can we delete this branch, or do you want to keep it around for future use/reference?

tomnez:
Nah we can balete it.

PR #324, opened 2019-06-05
Upgrade - use native ES6 class syntax with decorators

tomnez:
Ran https://github.com/ember-codemods/ember-es6-class-codemod and then fixed any issues that came up.

Tests are still being fixed in a separate PR at https://github.com/smile-io/ember-polaris/pull/322

**Note** This introduces breaking changes where we can no longer have component attributes named `action` while also using the `@action` decorator in the same file. So there are a few components where `action` has been renamed to `primaryAction`, which means any host apps that pull in the upgraded release will need to update their usages.


addon/components/polaris-action-list/section.js
```diff
@@ -1,13 +1,12 @@
+import { tagName, layout as templateLayout } from '@ember-decorators/component';
+import { action, computed } from '@ember/object';
```
andrewpye: Just wondering if it's possible to use `action as actionDecorator` (or something better-named 😅) to avoid having to rename every property that's called `action` - that way we might be able to avoid diverging from the React implementation's interface and eliminate the need for breaking changes... 🙏

vladucu: do we have many properties called `action`?

PS: I don't think we should make this decision based on whether we need to make a breaking change or not (that's why semver exists) or just to avoid diverging from React implementation - we already diverge in quite a few places. From my view, we should think about what would be best for the future of the add-on

andrewpye: >do we have many properties called `action`?

Not sure without checking tbh 🤷‍♂ 

I agree with the rest of what you're saying... but if there's a way to have both (i.e. use the `@action` decorator _and_ keep our existing naming) just by choosing a different name for the decorator, why would we choose a breaking change over an internal rename? 🤔

vladucu: > but if there's a way to have both (i.e. use the @action decorator and keep our existing naming) just by choosing a different name for the decorator, why would we choose a breaking change over an internal rename? 

not sure....kind of leaning towards ensuring we avoid risks of conflicting with Ember internals out-of-the-box in detriment of not matching React implementation 💯(we're already doing this in a few places, right?) - since while that's nice to have, not really a requirement....but I'm not super strong on this

addon/components/polaris-date-picker/month.js
```diff
@@ -9,152 +14,153 @@ import {
...
+  monthName /* month */() {}
 
   /**
    * @property weekdayName
    * @public
    * @type {Function}
    * @default noop
    */
-  weekdayName(/* weekday */) {},
+  weekdayName /* weekday */() {}
```
andrewpye: lol, arg comment got messed up here 🤷‍♂

sivakumar-kailasam: Use `/** weekday */` for commenting args. Using a single asterisk can throw off many tools

andrewpye: TIL 😄

addon/components/polaris-action-list.js
```diff
@@ -53,17 +52,19 @@ export default Component.extend({
...
    * @default no-op
    */
-  onActionAnyItem() {},
+  onActionAnyItem() {}
 
   /**
    * @private
    */
-  hasMultipleSections: gt('finalSections.length', 1).readOnly(),
+  @(gt('finalSections.length', 1).readOnly())
```
sivakumar-kailasam: wouldn't this be,
```javascript
@readOnly
@gt('finalSections.length', 1)
hasMultipleSections
```

addon/components/polaris-button.js
```diff
@@ -322,107 +319,109 @@ export default Component.extend({
...
+      outline,
+      destructive,
+      isDisabled,
+      loading,
+      plain,
+      monochrome,
+      size,
+      fullWidth,
+      isIconOnly,
+    } = this.getProperties(
```
sivakumar-kailasam: we could probably replace this with  `let { ... } = this;` instead of requiring `this.getProperties`

vladucu:
@tomnez there seems to be a few files on which the codemod likely failed and we'll have to try again
Screen Shot 2019-06-07 at 12 29 48 PM

addon/components/polaris-button.js
tomnez: There are a lot of instances of`get` and `getProperties` left, I feel like this is out of scope for running the codemod to get ES6 classes and decorators 🤔

sivakumar-kailasam: Sounds reasonable 🙂

addon/components/polaris-action-list.js
tomnez: ah yeah looks like the codemod also didn't set all the `@readOnly` decorators properly 😞

tomnez: update, looks like the codemod is doing the right thing. If trying to use `@readOnly` on its own, it errors with

```
You attempted to use @readOnly as a decorator directly, but it requires a `dependentKey` parameter
```

If I pass any keys into it, it errors with

```
Only one computed property decorator can be applied to a class field or accessor, but 'hasMultipleSections' was decorated twice.
```

🤷‍♂

andrewpye: > If trying to use @readOnly on its own, it errors

I guess it's trying to use the `readOnly` CP macro, instead of the modifier (same as the old `import { readOnly } from '@ember/object/computed'` 🤔 

In ES6-land, maybe these are defined with a getter and no setter, making them read-only anyway? Not sure...

tomnez: Yeah seems like it, and I don't see any `readOnly` available as an import from `ember-decorators` currently so I guess for now we have to stick with `.readOnly()`

andrewpye: Not much else we can do at this point by the sounds of it then 🤷‍♂

vladucu: yeah, I think Tom & the codemod are correct 😄 

PS: same old `import { readOnly } from '@ember/object/computed'` should work now as a decorator for ` readOnly` CP, but this is a `readOnly` CP modifier (or whatever they are called)

tomnez:
Added some new commits, the codemod doesn't play nice with files that have custom computed methods (such as `computedHelpTextId`, etc). We can still use them as decorators but I had to comment them out and _then_ run the codemod, and then go back and fix up the custom stuff manually.

vladucu:
> Added some new commits, the codemod doesn't play nice with files that have custom computed methods (such as computedHelpTextId, etc). We can still use them as decorators but I had to comment them out and then run the codemod, and then go back and fix up the custom stuff manually.

yeap, that's it tomster 😄 , we just have to comment out small bits and then fix manually 💪

vladucu:
great stuff in here
closing this in favor of #360 and using this as a reference

PR #325, opened 2019-06-21
Add the ability to disable shortcutActions in polaris resource list item

sivakumar-kailasam:


andrewpye:
@sivakumar-kailasam just noticed some test failures on CI, fix those up and you're good to go 👍

sivakumar-kailasam:
> Just thinking, places like this are where using angle-brackets and the spread operator would be super-handy so we don't have to pass every property through explicitly..

For sure. Let’s wait until octane is out & we can probably do this & a lot more with element modifiers to improve the DX of this addon.

The test failures seem unrelated to this change, but I'll add the fix.

vladucu:
> places like this are where using angle-brackets and the spread operator would be super-handy so we don't have to pass every property through explicitly..

sadly, current spread-operator only forwards HTML attributes, not custom component arguments 😞

sivakumar-kailasam:
I haven't tried `...{this.args}` but that might work 🙂

PR #358, opened 2019-08-07
Refactor polaris-button to ES6 class

vladucu:
### What's here

- refactors to ES6 class
- template-only components
- angle-bracket syntax
- deprecates `externalClasses` - going forward an HTML `class` attribute can be used

#### NOTE
Tests are failing on beta & canary, likely related to ember-decorators/ember-decorators#451


addon/components/polaris-button.js
```diff
@@ -322,107 +318,122 @@ export default Component.extend({
...
+      outline,
+      destructive,
+      isDisabled,
+      loading,
+      plain,
+      monochrome,
+      size,
+      fullWidth,
+      isIconOnly,
+    } = this.getProperties(
```
sivakumar-kailasam: we could probably skip using `getProperties` and directly destructure from `this`

addon/components/polaris-button.js
```diff
@@ -322,107 +318,122 @@ export default Component.extend({
...
+      classes.push('Polaris-Button--iconOnly');
     }
-  ).readOnly(),
 
-  type: computed('submit', function() {
+    return classes.join(' ');
+  }
+
+  @(computed('submit').readOnly())
+  get type() {
```
sivakumar-kailasam: unless the getters have corresponding setters, we can skip using readOnly. There are similar instances in this review.

addon/components/polaris-button.js
```diff
@@ -322,107 +318,122 @@ export default Component.extend({
...
-     * button element's event object as the first parameter.
-     */
-    invokeMouseAction(actionName) {
-      return this.get(actionName)();
-    },
-  },
-});
+  }
+
+  init() {
```
sivakumar-kailasam: init should be replaced with constructor

addon/components/polaris-button.js
vladucu: ohh, indeed, we support `3.10+`...will update

addon/components/polaris-button.js
vladucu: good catch, thx!

addon/components/polaris-button.js
vladucu: for components? do you have a source for this?
I was under the impression that it's recommended to still use `init` if extending from an `EmberObject` based class

sivakumar-kailasam: my bad. you're right 🙂I overlooked the fact that this is still an ember component and not a glimmer component.

vladucu: no worries...thinking of actually switching to a glimmer component while here

addon/components/polaris-button.js
```diff
@@ -322,107 +318,110 @@ export default Component.extend({
...
+
+  @computed('submit')
+  get type() {
+    return this.submit === true ? 'submit' : 'button';
+  }
 
-  ariaExpandedValue: computed('ariaExpanded', function() {
-    let ariaExpanded = this.get('ariaExpanded');
+  @computed('ariaExpanded')
+  get ariaExpandedValue() {
```
sivakumar-kailasam: this could probably go directly on the template. If an attribute is empty, it won't be rendered anyway.

addon/components/polaris-button.js
```diff
@@ -322,107 +318,110 @@ export default Component.extend({
...
 
-  type: computed('submit', function() {
-    return this.get('submit') === true ? 'submit' : 'button';
-  }).readOnly(),
+    return classes.join(' ');
+  }
+
+  @computed('submit')
+  get type() {
+    return this.submit === true ? 'submit' : 'button';
```
sivakumar-kailasam: how do you feel about moving this to the template to help shorten this class?

addon/components/polaris-button.js
```diff
@@ -322,107 +318,110 @@ export default Component.extend({
...
+      }
+    );
+  }
+
+  /**
+   * Helper to invoke passed-in actions without passing the
+   * button element's event object as the first parameter.
+   */
+  @action
+  invokeMouseAction(actionName) {
```
sivakumar-kailasam: I've never realised that we don't send back the event. What if someone wants the event? (to prevent bubbling)

vladucu:
good feedback Siva...updated

PR #359, opened 2019-08-07
Refactor polaris-drop-zone

vladucu:
### What's here

- refactors to ES6 class
- template-only components
- angle-bracket syntax

##### NOTE
Tests are failing on beta & canary, likely related to ember-decorators/ember-decorators#451

### DIFF
--- a/addon/components/polaris-drop-zone.js
+++ b/addon/components/polaris-drop-zone.js
@@ -1,14 +1,15 @@
 import Component from '@ember/component';
-import { computed } from '@ember/object';
+import { action, computed, setProperties } from '@ember/object';
 import { or } from '@ember/object/computed';
 import { classify } from '@ember/string';
 import { throttle, scheduleOnce } from '@ember/runloop';
 import { isNone, isPresent } from '@ember/utils';
+import { layout } from '@ember-decorators/component';
 import ContextBoundEventListenersMixin from 'ember-lifeline/mixins/dom';
 import { getRectForNode } from '@shopify/javascript-utilities/geometry';
 import { guidFor } from '@ember/object/internals';
-import layout from '../templates/components/polaris-drop-zone';
-import State from '../-private/drop-zone-state';
+import template from '../templates/components/polaris-drop-zone';
+import DropZoneState from '../-private/drop-zone-state';
 import {
   fileAccepted,
   getDataTransferFiles,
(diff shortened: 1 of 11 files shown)


addon/components/polaris-drop-zone.js
```diff
@@ -280,73 +283,67 @@ export default Component.extend(ContextBoundEventListenersMixin, {
...
-    'state.error',
-    'sizeClass',
-    function() {
-      let { outline, isDragging, state, sizeClass } = this.getProperties(
-        'outline',
-        'isDragging',
-        'state',
-        'sizeClass'
-      );
+  @computed()
```
tomnez: do we want to keep the `.readOnly()` on all these CPs?

addon/components/polaris-drop-zone.js
```diff
@@ -543,139 +510,124 @@ export default Component.extend(ContextBoundEventListenersMixin, {
...
+    this.getDerivedStateFromProps();
+  }
 
+  /**
+   * NOTE: if the component is rendered with `openFileDialog = true`, this will
+   * not work as expected: the file chooser is not opened because of how the HTML
+   * input element works. This should be a user activation.
+   *
+   * Following warning will be logged
+   * `File chooser dialog can only be shown with a user activation.`
```
tomnez: 👍

addon/templates/components/polaris-drop-zone.hbs
```diff
@@ -1,57 +1,58 @@
...
-    tagName=""
-  )
-) as |wrapper|}}
-  {{#component wrapper}}
+{{#let (if @label
+  (component "polaris-labelled")
+  (component "wrapper-element" tagName="")
+) as |Wrapper|}}
+  <Wrapper
+    @id={{this.id}}
```
tomnez: Should this be `@id` instead of `this.id` since `id` is a public [attribute](https://github.com/smile-io/ember-polaris/blob/43ea1c80acb2379b0cd29a5f7d2aa11ca3a1220a/addon/components/polaris-drop-zone.js#L37)?

addon/components/polaris-drop-zone.js
vladucu: no, having a computed property with only a getter....trying to set it, should throw an error

addon/templates/components/polaris-drop-zone.hbs
vladucu: yes indeed...good catch
fixed

PR #363, opened 2019-08-09
Refactor polaris-visually-hidden component

vladucu:
#### What's here

- refactors to ES6 class
- template-only components
- angle-bracket syntax


vladucu:
> Maybe use glimmer components for simple components like these?

Was thinking to leave that for later....tried it for a bit but tests were throwing some random errors....maybe I'll give another shot

> Also the fallback behavior won't work on templates since we're using @ args unless we implement that behavior on unit. Best we use glimmer and use the or helper to fallback to defaults

which fallback?

sivakumar-kailasam:
> which fallback?
the `text` property on the component class.

vladucu:
> the text property on the component class.

well that's `null`...should we care about these?

PR #369, opened 2019-08-12
Refactor  polaris-page

vladucu:
#### What's here

- refactors to ES6 class
- template-only components
- angle-bracket syntax

### DIFF
--- a/addon/components/polaris-page.js
+++ b/addon/components/polaris-page.js
@@ -1,121 +1,106 @@
 import Component from '@ember/component';
 import { or } from '@ember/object/computed';
-import layout from '../templates/components/polaris-page';
+import { tagName, layout } from '@ember-decorators/component';
+import template from '../templates/components/polaris-page';
 
 /**
  * Polaris page component.
  * See https://polaris.shopify.com/components/structure/page
  */
-export default Component.extend({
-  classNames: ['Polaris-Page'],
-  classNameBindings: [
-    'fullWidth:Polaris-Page--fullWidth',
-    'singleColumn:Polaris-Page--singleColumn',
-  ],
-
-  layout,
-
+@tagName('')
+@layout(template)
+export default class PolarisPageComponent extends Component {
   /**
    * Page title, in large type
    *
-   * @property title
-   * @public
    * @type {String}
    * @default null
+   * @public
    */
-  title: null,
+  title = null;
 
   /**
    * Important and non-interactive status information shown immediately after the title
    *
-   * @property titleMetadata
-   * @public
    * @type {String|Component}
    * @default null
+   * @public
    */
-  titleMetadata: null,
+  titleMetadata = null;
 
   /**
    * Visually hide the title (stand-alone app use only)
    *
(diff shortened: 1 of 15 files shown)


addon/templates/components/polaris-page/action.hbs
```diff
@@ -1,16 +1,29 @@
-{{!-- TODO: support url --}}
-{{#if (or icon disclosure)}}
-  <span class="Polaris-Header-Action__ActionContent">
-    {{#if icon}}
-      {{polaris-page/action-icon icon=icon}}
-    {{/if}}
+{{! TODO: support url }}
+<button
+  class={{concat
```
tomnez: Can we drop `concat` here?

```html
<button class="
  Polaris-Header-Action
  {{if this.isIconOnly "Polaris-Header-Action--iconOnly"}}
  {{if @disabled "Polaris-Header-Action--disabled"}}
">
  ...
</button>
```

addon/templates/components/polaris-page/header.hbs
```diff
@@ -1,120 +1,119 @@
...
-      {{/if}}
-    </div>
-  {{/if}}
-
-  {{!-- Add a "main content" div if we have a primary action. --}}
-  {{#wrapper-element
-    tagName=(if primaryAction "div" "")
-    class=(if primaryAction "Polaris-Page-Header__MainContent")
+<div
+  class={{concat
```
tomnez: and here?

vladucu:
bump @andrewpye

addon/components/polaris-page.js
```diff
@@ -1,121 +1,107 @@
...
-export default Component.extend({
-  classNames: ['Polaris-Page'],
-  classNameBindings: [
-    'fullWidth:Polaris-Page--fullWidth',
-    'singleColumn:Polaris-Page--singleColumn',
-  ],
-
-  layout,
-
+@tagName('')
```
andrewpye: Will this break any instances where we have `class` set on `polaris-page` instances? If so, do we have any such instances? Even if we don't, this would still be a breaking change and we should have that in mind when we release this update.

addon/templates/components/polaris-page/header.hbs
```diff
@@ -1,120 +1,110 @@
...
-      as |primaryActionButton|
-    }}
-      {{!-- Add a "title and actions" wrapper div if we have a primary action. --}}
-      {{#wrapper-element
-        tagName=(if primaryAction "div" "")
-        class=(if primaryAction "Polaris-Page-Header__TitleAndActions")
-      }}
+      (component "wrapper-element" tagName="")
+    )
+    as |Rollup MainContent TitleAndActions PrimaryActionButton|
```
andrewpye: `:obama:`

addon/components/polaris-page.js
vladucu: Good catch!

Yeah, totally forgot to add backwards support for that and a deprecation. Will add and double-check components already refactored, I know I added this to some, but most likely not all
👍

PR #385, opened 2019-09-09
Refactor polaris-stack

vladucu:
### What's here

- refactors to ES6 class
- template-only components
- angle-bracket syntax

NOTE: Spent a bit of time looking for a solution to our auto-wrapping thingy and ultimately landed on using `MutationObserver`s. This seems to work well.

However, in the long term I'm wondering if we should think about just having a codemod and make usage explicit everywhere? Will be much more verbose for sure.....so not sold on it
Curious what you're thoughts are


addon/templates/components/polaris-drop-zone/file-upload.hbs
```diff
@@ -15,9 +15,7 @@
     <PolarisStack @vertical={{true}} @spacing="tight">
       {{#if (eq this.type "file")}}
         <img class={{this.imageClasses}} src={{this.fileUpload}} alt="">
-      {{/if}}
-
-      {{#if (eq this.type "image")}}
+      {{else if (eq this.type "image")}}
```
andrewpye: lol, not quite sure how these ended up falling under the `polaris-stack` refactor 🙈😛

addon/templates/components/polaris-stack.hbs
```diff
@@ -1,9 +1,13 @@
...
+  {{did-insert this.setupAutoWrapper}}
+  {{will-destroy this.teardownAutoWrapper}}
+  ...attributes
+>
+  {{#if hasBlock}}
+    {{yield (hash item=(component "polaris-stack/item"))}}
+  {{else}}
+    {{text}}
+  {{/if}}
+</div>
```
andrewpye: Nit: missing newline

addon/templates/components/polaris-stack/item.hbs
```diff
@@ -1,5 +1,11 @@
...
+  data-test-stack-item={{true}}
+  class={{this.classes}}
+  ...attributes
+>
+  {{#if hasBlock}}
+    {{yield}}
+  {{else}}
+    {{text}}
+  {{/if}}
+</div>
```
andrewpye: Nit: missing newline

vladucu:
>  Not sure where I stand on this, which means that for now I'm good with leaving this as is if it still works 

👍

PR #410, opened 2019-10-15
Remove jquery dependency

sivakumar-kailasam:
- Replaced jquery usage in color picker's code with native equivalents
- Added a test page that has an example of a color picker
- Removed jquery dependency & disabled it in optional features
- Updated ember-try config.

Screenshot of the test page:
localhost_4200_color-picker


vladucu:
closing this as outdated

PR #422, opened 2019-11-11
Fix linting issues

vladucu:
Fixes bunch of linting issues on `master` likely after some recent upgrades


vladucu:
merging this, any comments welcome

PR #425, opened 2019-11-15
Refactor polaris-avatar component

vladucu:
### What's here

- refactors to ES6 class
- template-only components
- angle-bracket syntax

### DIFF
--- a/addon/components/polaris-avatar.js
+++ b/addon/components/polaris-avatar.js
@@ -3,14 +3,8 @@ import { action, computed } from '@ember/object';
 import { or } from '@ember/object/computed';
 import { isEmpty } from '@ember/utils';
 import { classify } from '@ember/string';
-import {
-  classNames,
-  attributeBindings,
-  classNameBindings,
-  tagName,
-  layout as templateLayout,
-} from '@ember-decorators/component';
-import layout from '../templates/components/polaris-avatar';
+import { tagName, layout } from '@ember-decorators/component';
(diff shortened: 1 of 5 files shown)


addon/templates/components/polaris-avatar.hbs
```diff
@@ -1,29 +1,36 @@
...
-        font-size="26"
-        text-anchor="middle"
-      >
-        {{initials}}
-      </text>
-    </svg>
-  </span>
-{{/if}}
+<span
+  role={{this.role}}
```
andrewpye: Any reason to not just hard-code this now? 🤔

addon/components/polaris-avatar.js
```diff
@@ -256,50 +221,58 @@ export default Component.extend({
...
-      'finalSource',
-      'hasError'
-    );
+  @computed('finalSource', 'hasError')
+  get shouldShowImage() {
+    let { finalSource, hasError } = this;
     return finalSource && !hasError;
-  }).readOnly(),
-
-  didReceiveAttrs() {
```
andrewpye: Why did this disappear?

vladucu: it's replaced by `resetImage` action below

addon/templates/components/polaris-avatar.hbs
vladucu: ohh, it's private property
yeah, we can do 👍

PR #447, opened 2019-11-29
Refactor polaris-button-group component

vladucu:


addon/components/polaris-button-group.js
```diff
@@ -1,80 +1,67 @@
...
-    // Wrap each child element that isn't already a group item.
-    let nodesToWrap = rejectNodesByClassName(
-      this.element.children,
-      'Polaris-ButtonGroup__Item'
-    );
-    let wrapper = document.createElement('div');
+  @action
+  setupAutoWrapper(element) {
+    // TODO When we're auto-wrapping we bypass the group item component, so we will not handle
+    // focus events and hence not apply correct class to the wrapper group item
```
andrewpye: Is this a new issue? When does it occur/what are the symptoms?

Also thinking... can we ever handle this without some (probably hacky) workaround? 🤔

addon/templates/components/polaris-resource-list/item.hbs
```diff
@@ -79,11 +79,11 @@
       class="Polaris-ResourceList-Item__Actions"
       click=(action stopPropagation)
     }}
-      {{#polaris-button-group
-        segmented=(not persistActions)
-        data-test-id=(unless persistActions "shortcut-actions")
+      <PolarisButtonGroup
+        data-test-id={{unless persistActions "shortcut-actions"}}
```
andrewpye:
```suggestion
        data-test-id={{unless @persistActions "shortcut-actions"}}
```

?

addon/components/polaris-button-group.js
vladucu: This is an existing issue....whenever users of this component are not explicitly wrapping buttons inside `buttonGroup.item` components, we will auto-wrap buttons with a `div` (what we do here), but that obviously will not handle focus events that the `buttonGroup.item` component does

Can't think of something nice here, only if we could easily render the item component inline and wrap with it contents....sounds crazy
I'd say it might worth just deprecating this auto-wrapping for components like this (not sure if there are other)

addon/templates/components/polaris-resource-list/item.hbs
vladucu: we will get to this component too shortly....this was just to get tests passing

addon/components/polaris-button-group.js
andrewpye: Yeah, probably removing auto-wrapping is the way to go to work around this, any other "fix" would probably have to be dirty 😬

PR #449, opened 2019-11-29
Modernize polaris-caption component

vladucu:


addon/templates/components/polaris-caption.hbs
```diff
@@ -1,5 +1,7 @@
-{{#if hasBlock}}
-  {{yield}}
-{{else}}
-  {{text}}
-{{/if}}
+<p class="Polaris-Caption" ...attributes>
```
andrewpye: Is this missing the `data-test-caption` attribute?

vladucu:
> Got some test failures on CI 😞
generally speaking, don't worry about this....will not merge to break the build 😉 

on it

PR #451, opened 2019-11-29
Modernize polaris-choice-list component

vladucu:


addon/components/polaris-choice-list.js
```diff
@@ -58,103 +48,84 @@ export default Component.extend({
...
    *
-   * @property choices
-   * @public
    * @type {Array}
-   * @default null
+   * @default []
+   * @public
    */
-  choices: null,
+  choices = [];
```
andrewpye: Any particular reason for changing this? Also if this has changed from `null` to `[]`, should we make the same change to `selected` for consistency? 🤔

vladucu: changed this because it should be the default....we didn't do this before because of leaking issue before refactored into an ES6

yeah, can make selected too 👍

PR #493, opened 2020-02-05
Modernize polaris-color-picker components

tomnez:
Updates `PolarisColorPicker` to tagless component, ES6 classes, and angle bracket syntax.

A lot of changes are just codemod changes - additional edits to the component worth pointing out are:

- Use of `{{did-insert}}` and `{{will-destroy}}` modifiers so we can set a unique ID on the tagless component since the component files made use of `this.element`.
- Added a `get element()` computed property to return the container element, if we think this will be confusing we could rename to something specific per component, i.e. `get alphaPickerElement()` 
- Added `template-lint-disable no-invalid-interactive` to the slider because it makes use of `onmousedown` and `ontouchstart` events.

#### Dummy app demo:

picker


addon/components/polaris-color-picker.js
```diff
@@ -63,97 +64,117 @@ export default Component.extend({
...
     const {
       color: { brightness },
       pickerSize,
     } = this.getProperties('color', 'pickerSize');
     return clamp(pickerSize - brightness * pickerSize, 0, pickerSize);
-  }).readOnly(),
+  }
+
+  @computed('contentId')
+  get element() {
```
andrewpye: Curious what @vladucu's thoughts are about this naming, mostly because I'm conflicted: part of me feels like we shouldn't be overwriting Ember's internal API, part of me feels like we _should_ be able to overwrite because _that's why we do inheritance_. Basically I'm good with this as is but would also understand if we'd prefer to name it something else 😝

addon/components/polaris-color-picker.js
```diff
@@ -63,97 +64,117 @@ export default Component.extend({
...
     if (isNone(mainColorElement)) {
       return;
     }
 
     this.set('pickerSize', mainColorElement.clientWidth);
-  },
+  }
+
+  @action
+  setContentId(element, [value]) {
```
andrewpye: Reckon this, then `contentId` property etc. are worth putting into a mixin? We're already using them 4 times in this PR alone 🤔

addon/components/polaris-color-picker.js
```diff
@@ -63,97 +64,117 @@ export default Component.extend({
...
+      saturation,
+      brightness,
+      alpha,
+    });
+  }
 
-      const saturation = clamp(x / pickerSize, 0, 1);
-      const brightness = clamp(1 - y / pickerSize, 0, 1);
+  @action
+  handleHueChange(hue) {
```
andrewpye: huehuehue

addon/components/polaris-color-picker.js
```diff
@@ -1,5 +1,7 @@
+import { tagName, layout as templateLayout } from '@ember-decorators/component';
+import { action, computed } from '@ember/object';
+import { guidFor } from '@ember/object/internals';
 import Component from '@ember/component';
```
vladucu: can we please order this a bit as we usually do?

addon/components/polaris-color-picker.js
vladucu: hmm, tbh...I'm not sure I understand why we need this

addon/components/polaris-color-picker.js
```diff
@@ -63,97 +64,117 @@ export default Component.extend({
...
+  }
+
+  @computed('contentId')
+  get element() {
+    return document.querySelector(`#${this.contentId}`);
+  }
 
   didRender() {
-    this._super(...arguments);
+    super.didRender(...arguments);
```
vladucu: I don't think we need `didRender` anymore...actually we want to stop using any other lifecycle hooks except init/destroy....instead we should have this logic inside the action you added `setContentId` (renamed ofc)

addon/components/polaris-color-picker.js
```diff
@@ -63,97 +64,117 @@ export default Component.extend({
...
+    this.set('contentId', value);
+  }
+
+  @action
+  draggerMoved({ x, y }) {
+    const {
+      pickerSize,
+      color: { hue, alpha = 1 },
+      onChange,
+    } = this.getProperties('pickerSize', 'color', 'onChange');
```
vladucu:
```suggestion
    } = this;
```

addon/components/polaris-color-picker.js
```diff
@@ -63,97 +64,117 @@ export default Component.extend({
...
+  }
 
-      const saturation = clamp(x / pickerSize, 0, 1);
-      const brightness = clamp(1 - y / pickerSize, 0, 1);
+  @action
+  handleHueChange(hue) {
+    const {
+      color: { brightness, saturation, alpha = 1 },
+      onChange,
+    } = this.getProperties('color', 'onChange');
```
vladucu:
```suggestion
    } = this;
```

addon/components/polaris-color-picker.js
```diff
@@ -63,97 +64,117 @@ export default Component.extend({
...
+      });
+    }
+  }
+
+  @action
+  handleAlphaChange(alpha) {
+    const {
+      color: { hue, brightness, saturation },
+      onChange,
+    } = this.getProperties('color', 'onChange');
```
vladucu:
```suggestion
    } = this;
```

addon/components/polaris-color-picker/alpha-picker.js
```diff
@@ -1,6 +1,8 @@
+import { tagName, layout as templateLayout } from '@ember-decorators/component';
```
vladucu: same here (and other files), can we order this please

addon/components/polaris-color-picker/alpha-picker.js
```diff
@@ -23,77 +25,91 @@ function alphaForOffset(offset, sliderHeight) {
...
       'sliderHeight',
       'draggerHeight'
     );
     const offset = offsetForAlpha(alpha, sliderHeight, draggerHeight);
     return clamp(offset, 0, sliderHeight);
-  }).readOnly(),
+  }
+
+  @computed('contentId')
+  get element() {
```
vladucu: similar here like in main component

addon/components/polaris-color-picker/alpha-picker.js
```diff
@@ -23,77 +25,91 @@ function alphaForOffset(offset, sliderHeight) {
...
-
-      onChange(alpha);
-    },
-  },
-});
+  }
+
+  @action
+  handleChange({ y }) {
+    const { sliderHeight, onChange } = this.getProperties(
```
vladucu: let's use ES6 destructuring

addon/components/polaris-color-picker.js
tomnez: @vladucu you're confused as to why it's a CP, or confused as to why we need to set the element?

I guess we could just set it in the init hook since that's the only place it's used. I was just trying to keep the same API that we were used to (having the use of `this.element`) but I can get rid of it. If we ever end up needing it in multiple methods then we can refactor to a CP again.

vladucu: yeah, why we need the CP....we get the `element` explicitly as first parameter in the action called through `did-insert` modifier....we could just set it here as a property on the component....but unless I'm wrong, it's only used in this one place

tomnez: update - was going through other files, and components like `polaris-color-picker/slidable` call `this.element` from multiple places, so maybe it's easiest to have it as a CP in this particular case?

If not we can call `document.querySelector('#${this.contentId}')` each time. Pretty much the same thing IMO 🤷‍♂

vladucu: note: maybe I'm missing something and just talking non-sense...feel free to correct me
what I mean is....we can just keep it simple....why add a CP when you already have an action that set's `contentId` and then have the CP compute what you call `element` now

even more...could this  just be? Not sure where the `content` stuff comes from in this PR....
```js

/**
 * Polaris color picker component.
 * See https://polaris.shopify.com/components/forms/color-picker
 */
@tagName('')
@templateLayout(layout)
export default class PolarisColorPicker extends Component {
  ....

// NO LONGER NEEDED
//  didRender() {
//    super.didRender(...arguments);

//    let element = this.element;

//    if (isNone(element)) {
//      return;
//    }

    // Grab the size of the picker for positioning the draggable markers.
//    const mainColorElement = element.querySelector(
//      'div.Polaris-ColorPicker__MainColor'
//    );
//    if (isNone(mainColorElement)) {
//      return;
//    }

//    this.set('pickerSize', mainColorElement.clientWidth);
//  }

  @action
  setPickerSize(element) {
    // Grab the size of the picker for positioning the draggable markers.
    const mainColorElement = element.querySelector(
      'div.Polaris-ColorPicker__MainColor'
    );
    this.set('pickerSize', mainColorElement.clientWidth);
  }

  ...
}
```

```hbs
<div
  id={{this.contentId}}
  class="Polaris-ColorPicker"
  {{did-insert this.setContentId}}
>
  <div class="Polaris-ColorPicker__MainColor">
    ...
  </div>
  ....
</div>
```

tomnez: ^ Agreed, this works for `polaris-color-picker`, but other components will have multiple methods that need to access the container element, so in other components we should at least set `element` as a property on `did-insert` (see `slidable.js`, it accessing `element` in other places besides the `did-insert` method)

vladucu: don't get me wrong Tom....in the same action you call from `did-insert`, of course, keep a reference to the `mainColorElement` or whatever you need by setting it on the component

tomnez:
@vladucu @andrewpye updated - also, I ran the base branch locally and `ember-beta` and `ember-canary` are having issues there as well. What are your thoughts on just removing those from our ember-try config while we get the rest of these components updated?

Seems like it would allow us to move a lot faster rather than finding failing tests and `skip`ping them.

addon/components/polaris-color-picker/slidable.js
```diff
@@ -142,19 +110,32 @@ export default Component.extend({
...
-  },
+  }
+
+  @computed('contentId')
+  get element() {
+    return document.querySelector(`#${this.contentId}`);
+  }
+
+  @action
+  setContentId(element, value) {
```
vladucu: is this copied from somewhere? I don't understand where it's coming from, can't see it in the version prior to these changes

tomnez: It's slighty different than the old version, it used to be

`setContentId(element, [value])`

because it used to get called from the `on-insert` modifier, but now it gets called from within `setContentIdAndTriggerHeightChanged` below.

vladucu: would you mind linking me to the old version here on github?

tomnez: It's here in my first commit - this link doesn't seem to scroll to the line changes, but its `slidable.js`, lines 150-153 - https://github.com/smile-io/ember-polaris/pull/493/commits/0c66046427b0f0b5bf6dfe02813057b85697560b#diff-160b5cb0d5fd7505911f238ca1295ee4R150-R153

vladucu: kk....do you think this is needed? why are we adding an ID attirbute to the new wrapping `div` in the template that was added after making this component tag-less?

tomnez: The component has other methods like `handleDraggerMove` which need to grab the main element, so we need some way of grabbing it. I left the `element` CP in here as [mentioned](https://github.com/smile-io/ember-polaris/pull/493#discussion_r375989667) earlier so other methods could `.get` the element when needed.

I'm open to doing it differently but we do need some way to access the main element.

vladucu: this should be much simpler....please check [ember-render-modifiers](https://github.com/emberjs/ember-render-modifiers) to understand it's usage....it's super simple: `did-insert` will invoke your action with one param `element` (the element on which `did-insert` is added) and whatever else you're passing to it.....just set this `element` argument you receive in your action and set it on this component

there's no need `guidFor` and appending `content`....etc (which is why I was confused)

tomnez: Ahh I did that originally, but got the error

`Cannot set property element of [object Object] which has only a getter`

I thought maybe it was something weird with the modifiers but it turns out it just doesn't want me using the key name `"element"`.. 🤦‍♂ So yeah I can just do that.

vladucu: welp, `element` is a reserved word in ember so makes sense

tomnez: Yeah, strange that it let me overwrite it with a CP tho.

andrewpye: > Yeah, strange that it let me overwrite it with a CP tho.

I guess overwriting replaces the entire setter/getter combo, while setting tries to access a non-existent setter on the existing `element` property 🤷‍♂

tomnez:
Ahhh yeah forgot to push 🤦‍♂

sivakumar-kailasam:
classic tom 😄

tomnez:
_Rough week_

addon/templates/components/polaris-color-picker.hbs
```diff
@@ -1,21 +1,26 @@
...
-  <div class="Polaris-ColorPicker__ColorLayer" style={{colorLayerStyle}}></div>
-  {{polaris-color-picker/slidable
-    draggerX=draggerX
-    draggerY=draggerY
-    onChange=(action "draggerMoved")
-  }}
-</div>
+<div
+  class="Polaris-ColorPicker"
+  {{did-insert this.setPickerSize}}
```
vladucu: nit: let's add `...attributes` here too

PR #494, opened 2020-02-07
Modernize polaris-data-table components

tomnez:
[DEV-56](https://smileio.atlassian.net/secure/RapidBoard.jspa?rapidView=6&modal=detail&selectedIssue=DEV-56)

Updates `PolarisDataTable` to tagless component, ES6 classes, and angle bracket syntax.

Only real notable change here was removing the `didInsertElement` and `didUpdateAttrs` hooks from the component file in favor of using modifiers in the template file.

Dummy app:

Screen Shot 2020-02-07 at 9 59 35 AM


addon/components/polaris-data-table.js
```diff
@@ -1,525 +1,490 @@
...
+  }
 
-      this.addEventHandlers();
-    },
+  init() {
+    super.init(...arguments);
 
-    didUpdateAttrs() {
-      this._super(...arguments);
+    this.setProperties({
```
vladucu: let's move these initializations where we declare the properties at the top of the file....it's now safe to do that 🙌

addon/components/polaris-data-table.js
```diff
@@ -1,525 +1,490 @@
...
+      heights: [],
+      preservedScrollPosition: {},
+    });
+  }
 
-      if (isEqual(this.get('oldAttrs'), this.get('attrs'))) {
-        return;
-      }
+  @action
+  dataTableDidInsert(element) {
```
vladucu: nit: `initDataTable` or `insertDataTable`?

addon/components/polaris-data-table.js
```diff
@@ -1,525 +1,490 @@
...
+    this.set('dataTableElement', element);
+    this.handleResize();
+    this.addEventHandlers();
+  }
 
-      this.handleResize();
+  @action
+  dataTableDidUpdate() {
```
vladucu: same re naming tweak

we need to be careful here and likely check Polaris in what circumstances they do these updates...because currently this action will only fire once (we'll have to pass the properties that might change and need to re-trigger this action in the template_

addon/templates/components/polaris-data-table.hbs
```diff
@@ -1,77 +1,84 @@
...
-                totalsRowHeading=totalsRowHeading
-              }}
+            {{#each @headings as |heading headingIndex|}}
+              <PolarisDataTable::Heading
+                @heading={{heading}}
+                @headingIndex={{headingIndex}}
+                @truncate={{@truncate}}
+                @heights={{this.heights}}
+                @sortable={{@sortable}}
+                @sortedColumnIndex={{or this.sortedColumnIndex this.initialSortColumnIndex}}
```
vladucu: is `initialSortColumnIndex` public?

tomnez: Ah good catch 👀

addon/components/polaris-data-table.js
tomnez: @vladucu looks like the only _public_ properties that get used in the `componentDidUpdate` method in the 3.10 DataTable are `footerContent` and `truncate`, so we can probably just pass those two properties into `did-update`.

The React version `return`s inside ``componentDidUpdate`` if `isEqual(prevProps, this.props)`, so I'm guessing we don't need to worry about any of our privately set properties needing to trigger an update.

PR #501, opened 2020-02-10
Modernize polaris-range-slider component

tomnez:
[DEV-77](https://smileio.atlassian.net/secure/RapidBoard.jspa?rapidView=6&modal=detail&selectedIssue=DEV-77)

Updates `PolarisRangeSlider` to tagless component, ES6 classes, and angle bracket syntax.

Dummy app:

range

^ I'm not sure what's going on with the positioning here but this is also happening on `master` but not in any apps that actually use `ember-polaris`, so I'm guessing it's just something weird with the way styles are being pulled from Polaris into the dummy app.

### DIFF
--- a/addon/templates/components/polaris-range-slider.hbs
+++ b/addon/templates/components/polaris-range-slider.hbs
@@ -1,6 +1,6 @@
 <PolarisLabelled
-  @dataTestLabelled={{dataTestRangeSlider}}
-  @id={{@id}}
+  @dataTestLabelled={{this.dataTestRangeSlider}}
+  @id={{this.id}}
   @label={{@label}}
   @error={{@error}}
   @action={{@labelAction}}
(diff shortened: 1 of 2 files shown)


addon/templates/components/polaris-range-slider.hbs
```diff
@@ -1,20 +1,20 @@
...
-  id=id
-  label=label
-  error=error
-  action=labelAction
-  labelHidden=labelHidden
-  helpText=helpText
-}}
+<PolarisLabelled
+  @dataTestLabelled={{this.dataTestRangeSlider}}
+  @id={{this.id}}
```
tomnez: @vladucu @andrewpye Is there a preferred way we want to fall back onto defaults for public attributes like this?

Right now I'm just using `this.` instead of `@`, but was thinking we might want to to do something like:

```
@id={{or @id this.id}}
```

for now during this transitional period just to be super explicit that we're falling back on a default? Same goes for `min`, `max`, `onFocus`, and `onBlur`.

vladucu: 👍 
Love that...nice idea

andrewpye:
> I'm not sure what's going on with the positioning here but this is also happening on `master` but not in any apps that actually use `ember-polaris`, so I'm guessing it's just something weird with the way styles are being pulled from Polaris into the dummy app.

Yeah this is the case with a bunch of things in the dummy app for some reason I never got around to figuring out 🤷‍♂ As long as these things work in real apps then I'm not that fussed...

PR #502, opened 2020-02-12
Modernize polaris-resource-list components

tomnez:
[DEV-78](https://smileio.atlassian.net/secure/RapidBoard.jspa?rapidView=6&modal=detail&selectedIssue=DEV-78)

Updates `PolarisResourceList` and all sub-components to tagless component, ES6 classes, and angle bracket syntax.

Notable changes:
- Removed `wrapperId` from the main `polaris-resource-list` component. It was only used as a selector in order to set the `listNode` attribute (`ul.Polaris-ResourceList`). This was breaking in tests because the `ul` wasn't rendered yet when setting the node in the component file, so instead I added a `{{did-insert this.insertListNode}}` on the `ul` itself and set the reference directly in there.
- Used the `(element)` template helper in favor of our `WrapperElement` component wherever I could.

Dummy app:

list

### DIFF
--- a/addon/components/polaris-resource-list.js
+++ b/addon/components/polaris-resource-list.js
@@ -1,9 +1,10 @@
 import Component from '@ember/component';
-import { computed, get } from '@ember/object';
-import { guidFor } from '@ember/object/internals';
 import { gt } from '@ember/object/computed';
+import { get, computed } from '@ember/object';
+import { guidFor } from '@ember/object/internals';
 import { htmlSafe } from '@ember/string';
 import { assert } from '@ember/debug';
+import { tagName, layout as templateLayout } from '@ember-decorators/component';
 import ContextBoundEventListenersMixin from 'ember-lifeline/mixins/dom';
 import ContextBoundTasksMixin from 'ember-lifeline/mixins/run';
 import createContext from 'ember-context';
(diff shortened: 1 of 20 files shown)


tomnez:
Working on the test failures, FYI

andrewpye:
@tomnez I'll give this a look over once tests are fixed up 👍

tomnez:
@andrewpye behold the passing tests 👼

addon/components/polaris-resource-list/filter-control/filter-value-selector.js
```diff
@@ -123,8 +118,17 @@ export default Component.extend({
...
-});
+  }
+
+  init() {
+    super.init(...arguments);
+
+    // Using `afterRender` here instead of `didInsertElement` hook.
+    // This component is tagless and has no container div so we can't
+    // use `{{did-insert}}` modifiers in the template.
+    scheduleOnce('afterRender', this, this.insertFilterValueSelector);
```
tomnez: I wanted to specifically point out this change since it feels kind of like a hax, but we get double-render errors since this method (the contents of `insertFilterValueSelector`) was originally moved from `didInsertElement` to `init`. We need to wait until the element is rendered to run this and this feels like the best way to achieve a `didInsertElement` without actually using the hook 😬 

This is tagless and the main blocks of the template are `<PolarisStack>`s, so we can't have `{{did-insert}}` modifiers on them because they're components (it works, but tests will error since a modifier on a component is not really acceptable and a bad practice).

addon/components/polaris-resource-list.js
```diff
@@ -28,809 +28,677 @@ const LARGE_SPINNER_HEIGHT = 45;
...
-
-    debouncedHandleResize() {
-      // In the React implementation, the resize event listener
-      // is only rendered when `selectable` is truthy. We handle
-      // that by bombing out of this handler when that condition
-      // is met rather than dynamically managing event listeners.
-      if (!this.get('selectable')) {
-        return;
-      }
+  @(computed('selectable', 'sortOptions.length', 'alternateTool').readOnly())
```
andrewpye: Do we need the `.readOnly()` here (and in some other places) if this CP is defined without a setter? 🤔

addon/components/polaris-resource-list/filter-control/date-selector.js
```diff
@@ -128,9 +127,9 @@ export default Component.extend({
    * @default null
    * @private
    */
-  initialConsumerFilterKey: null,
+  initialConsumerFilterKey = null;
 
-  dateComparatorOptions: computed(function() {
+  get dateComparatorOptions() {
```
andrewpye: Can this just be defined as a regular property now?

addon/components/polaris-resource-list/filter-control/date-selector.js
```diff
@@ -183,35 +182,27 @@ export default Component.extend({
...
+        ...this.datePastOptions,
+        ...this.dateFutureOptions,
+        ...this.dateComparatorOptions,
       ],
     };
-  }).readOnly(),
+  }
 
-  now: computed(function() {
+  get now() {
```
andrewpye: Does this value get memoized the same way the CP used to? 🤔

addon/components/polaris-resource-list/filter-control/filter-value-selector.js
andrewpye: Interesting, thanks for the explanation @tomnez! Maybe worth adding as a comment, unless @vladucu has a suggestion for getting around this?

addon/templates/components/polaris-resource-list/bulk-actions/bulk-action-button.hbs
```diff
@@ -1,49 +1,37 @@
...
-
-    {{#if disclosure}}
-      <span class="Polaris-ResourceList-BulkActions__ActionIcon">
-        {{polaris-icon source="caret-down"}}
-      </span>
-    {{/if}}
-  {{/component}}
+  {{#let (element (if @disclosure "span" "")) as |OuterWrapper|}}
+    <OuterWrapper
+      class="Polaris-ResourceList-BulkActions__ActionContent"
```
andrewpye: Just wanna check that this doesn't have any weird behaviours if we pass a `class` to `OuterWrapper` when it's tagless - guessing it just silently does nothing 🤷‍♂

addon/templates/components/polaris-resource-list/bulk-actions/bulk-action-button.hbs
```diff
@@ -1,49 +1,37 @@
...
-      {{else}}
-        {{render-content content}}
-      {{/if}}
-    {{/wrapper-element}}
-
-    {{#if disclosure}}
-      <span class="Polaris-ResourceList-BulkActions__ActionIcon">
-        {{polaris-icon source="caret-down"}}
-      </span>
-    {{/if}}
```
andrewpye: Have we lost this block somehow? 😕

addon/templates/components/polaris-resource-list/bulk-actions/bulk-action-button.hbs
```diff
@@ -1,49 +1,37 @@
...
-  {{/component}}
+  {{#let (element (if @disclosure "span" "")) as |OuterWrapper|}}
```
vladucu: 😍

addon/templates/components/polaris-resource-list/checkable-button.hbs
```diff
@@ -1,9 +1,21 @@
-<div class="Polaris-ResourceList-CheckableButton__Checkbox">
-  {{polaris-checkbox
-    labelHidden=true
-    label=accessibilityLabel
-    checked=selected
-    disabled=disabled
-  }}
+{{!-- template-lint-disable no-invalid-interactive --}}
+<div
+  {{on "click" this.onToggleAll}}
```
vladucu: let's follow what you did in other places
```suggestion
  {{on "click" (fn (or @onToggleAll this.onToggleAll))}}
```

addon/components/polaris-resource-list/filter-control/date-selector.js
vladucu: yeap, I think we can....nice catch 👍

addon/components/polaris-resource-list/filter-control/date-selector.js
vladucu: not without a `@computed` decorator

addon/components/polaris-resource-list/filter-control/filter-value-selector.js
vladucu: @tomnez I don't think it's that bad what you have here
one thought we could check....is to just have a simple `<span {{did-insert ...}}></span>` element added (the span should not cause any spacing issues I think)

addon/templates/components/polaris-resource-list/item.hbs
```diff
@@ -1,140 +1,162 @@
...
-              label=label
-              labelHidden=true
-              checked=isSelected
-              disabled=loading
-              onChange=(action handleSelection)
-            }}
-          {{/wrapper-element}}
-        {{/wrapper-element}}
-      {{/if}}
+  {{#with (if this.loading "-1" "0") as |tabIndex|}}
```
vladucu: let's use `let`

addon/templates/components/polaris-resource-list/item.hbs
```diff
@@ -1,140 +1,162 @@
...
+        class="Polaris-ResourceList-Item__Link"
+        @ariaDescribedBy={{@itemId}}
+        @ariaLabel={{@accessibilityLabel}}
+        @url={{@url}}
+        @tabIndex={{tabIndex}}
+        @onFocus={{fn this.handleAnchorFocus}}
+        @onBlur={{fn this.handleFocusedBlur}}
+      />
+    {{else}}
+      {{#let (element "button") as |ItemButton|}}
```
vladucu: do we need this `let` here? or can we just use the `button` directly below

addon/templates/components/polaris-resource-list/item.hbs
```diff
@@ -1,140 +1,162 @@
...
+        <ItemButton
+          class="Polaris-ResourceList-Item__Button"
+          aria-label={{@accessibilityLabel}}
+          aria-controls={{@ariaControls}}
+          aria-expanded={{@ariaExpanded}}
+          tab-index={{tabIndex}}
+          onclick={{fn this.handleClick}}
```
vladucu: if this is just a plain `button` element, can we use `{{on "click" ...}}` modifier here

addon/templates/components/polaris-resource-list/item.hbs
```diff
@@ -1,140 +1,162 @@
...
-              onClick=(action
-                (or shortcutAction.onAction (action (mut dummy)))
-              )
-            }}
-          {{/buttonGroup.item}}
-        {{/each}}
-      </PolarisButtonGroup>
-    {{/wrapper-element}}
+    {{#if (and @shortcutActions (not this.loading))}}
+      {{#let (element "div") as |ItemActions|}}
```
vladucu: again, for these other `let`'s....can't see a reason why we're doing this....unless I'm wrong and missing something let's just use the elements directly
cc @andrewpye in case you remember why we do this

addon/components/polaris-resource-list/filter-control/date-selector.js
andrewpye: Hmm OK... so we probably want dat `@computed` decorator here then to maintain behaviour

addon/components/polaris-resource-list/filter-control/filter-value-selector.js
tomnez: Yeah that works too 👍I wasn't sure if we wanted to go adding extra nodes that weren't in the React version but I'd rather use the `span` if we're all cool with it. Thanks ya'll.

addon/components/polaris-resource-list.js
tomnez: Hm this was all codemod magic, we've been keeping them in other branches. @vladucu any strong feelings on dropping `readOnly` or not for CP's without setters?

addon/components/polaris-resource-list/filter-control/date-selector.js
tomnez: 🤦‍♂ 👍

addon/components/polaris-resource-list.js
vladucu: I have no strong opinion here.....I have not commented on these in any PRs, just for the sake of getting these tasks done asap

we'll do some more tweaks at some point in the near future if we want to move to glimmer components with the new tracking feature and we can do at that point too

addon/templates/components/polaris-resource-list/bulk-actions/bulk-action-button.hbs
tomnez: Yeah I tested manually and it just doesn't create the span or a class.

addon/templates/components/polaris-resource-list/bulk-actions/bulk-action-button.hbs
tomnez: What the dang heck. Not sure what happened, will add back 😳

addon/templates/components/polaris-resource-list/item.hbs
tomnez: Hm yeah I was just replacing all `wrapper-element`s with the element helper.. but you're right I can't think of why we would have been using `wrapper-element` here in the first place since the `tagName` never changes 🤔Will see if @andrewpye can chime in here but seems like we can drop them now.

addon/components/polaris-resource-list/bulk-actions.js
```diff
@@ -1,462 +1,375 @@
...
+    this.onSelectModeToggle(val);
+  }
 
-        if (measuring) {
-          this.get('promotedActionsWidths').pushObject(width);
-        }
-      },
-    },
+  @action
+  handleMeasurement(width) {
```
vladucu: nice....that previous destructuring was really useless 👏

addon/templates/components/polaris-resource-list/bulk-actions.hbs
```diff
@@ -1,66 +1,66 @@
...
+            />
+          </popover.content>
+        </PolarisPopover>
       </div>
     {{/if}}
 
     <button
       class="Polaris-ResourceList-BulkActions__Button Polaris-ResourceList-BulkActions__Button--cancel"
-      onclick={{action "setSelectMode" false}}
+      onclick={{fn this.setSelectMode false}}
```
vladucu: can we use `on` modifier here ?

addon/templates/components/polaris-resource-list/filter-control/filter-value-selector.hbs
```diff
@@ -1,41 +1,49 @@
...
-    dateOptionType=filter.dateOptionType
-    filterValue=value
-    filterKey=filterKey
-    filterMinKey=filter.minKey
-    filterMaxKey=filter.maxKey
-    onFilterValueChange=(action onChange)
-    onFilterKeyChange=(action onFilterKeyChange)
-  }}
-{{/if}}
+<span {{did-insert (fn this.insertFilterValueSelector)}}>
```
vladucu: nit: no need for passing `fn` everywhere, only if you need to curry arguments

addon/templates/components/polaris-resource-list/item.hbs
vladucu: I think you missed this comment ^

addon/templates/components/polaris-resource-list/item.hbs
```diff
@@ -1,140 +1,162 @@
...
-  {{/if}}
+  <div
+    data-test-id="item-content"
+    class="Polaris-ResourceList-Item__Container"
+    id={{@itemId}}
+  >
+    {{#if (or @media this.selectable)}}
+      <div class="Polaris-ResourceList-Item__Owned">
+        {{#if this.selectable}}
+          {{#let (element "div") as |ItemHandle|}}
```
vladucu: same....all these `let (element...` where there's no condition....we can just use the element tag directly no?

addon/templates/components/polaris-resource-list/item.hbs
```diff
@@ -1,140 +1,162 @@
...
+            <ItemHandle
+              data-test-id="larger-selection-area"
+              class="Polaris-ResourceList-Item__Handle"
+              onclick={{fn this.handleLargerSelectionArea}}
```
vladucu: `on` render modifier

addon/templates/components/polaris-resource-list/item.hbs
```diff
@@ -1,140 +1,162 @@
...
+            >
+              {{#let (element "div") as |CheckboxWrapper|}}
+                <CheckboxWrapper
+                  class="Polaris-ResourceList-Item__CheckboxWrapper"
+                  onclick={{fn this.stopPropagation}}
```
vladucu: same, we can use `div` directly here and use `on` modifier

tomnez:
Thanks guys! 💪

PR #503, opened 2020-02-14
Modernize polaris-pagination component

tomnez:
[DEV-73](https://smileio.atlassian.net/secure/RapidBoard.jspa?rapidView=6&modal=detail&selectedIssue=DEV-73)

Updates `PolarisPagination` and all sub-components to tagless component, ES6 classes, and angle bracket syntax.

Not really demoable from the dummy app, it was two buttons next to each other, and the icons don't render anyway so it's not much to look at 😛

### DIFF
--- a/addon/components/polaris-pagination.js
+++ b/addon/components/polaris-pagination.js
@@ -1,23 +1,16 @@
 import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { not, and } from '@ember/object/computed';
 import { isArray } from '@ember/array';
 import { assert } from '@ember/debug';
-import { computed } from '@ember/object';
-import { and, not } from '@ember/object/computed';
 import { isEmpty } from '@ember/utils';
+import { tagName, layout as templateLayout } from '@ember-decorators/component';
 import layout from '../templates/components/polaris-pagination';
 import { handleMouseUpByBlurring } from '../utils/focus';
 
-export default Component.extend({
-  tagName: 'nav',
-
-  attributeBindings: ['accessibilityLabel:aria-label'],
-
-  classNames: ['Polaris-Pagination'],
-
-  classNameBindings: ['plain:Polaris-Pagination--plain'],
-
-  layout,
-
+@tagName('')
+@templateLayout(layout)
+export default class PolarisPagination extends Component {
   /**
    * A more subdued control for use in headers
    *
@@ -26,7 +19,7 @@ export default Component.extend({
    * @default false
    * @public
    */
-  plain: false,
+  plain = false;
 
   /**
    * The URL of the next page
@@ -37,7 +30,7 @@ export default Component.extend({
    * @default null
    * @public
    */
-  nextUrl: null,
+  nextUrl = null;
 
   /**
    * The URL of the previous page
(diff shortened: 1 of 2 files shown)


addon/components/polaris-pagination.js
```diff
@@ -118,56 +111,58 @@ export default Component.extend({
...
-  isNextDisabled: not('hasNext').readOnly(),
+  @(not('hasNext').readOnly())
+  isNextDisabled;
 
   /** @private */
-  hasNextKeysConfigured: computed('nextKeys.[]', function() {
-    if (isEmpty(this.nextKeys)) {
+  @(computed('nextKeys.[]').readOnly())
+  get hasNextKeysConfigured() {
+    let { nextKeys } = this;
```
vladucu: no need to change this, but I'm not sure, generally speaking, it's worth to have this declaration when it's used in 1-2 places and especially if the attr name is short (doesn't makes code harder to read)

addon/templates/components/polaris-pagination.hbs
```diff
@@ -1,39 +1,39 @@
...
+  ...attributes
 >
-  {{polaris-icon source="arrow-left"}}
-</button>
+  <button
+    data-test-prev-btn="true"
+    class="Polaris-Pagination__Button"
+    aria-label="Previous"
+    disabled={{this.isPreviousDisabled}}
+    onclick={{fn (or @onPrevious this.onPrevious)}}
```
vladucu: same...let's use `on` modifier ([here](https://ember-learn.github.io/ember-octane-vs-classic-cheat-sheet/)'s a nice cheat-sheet you can reference)

addon/templates/components/polaris-pagination.hbs
```diff
@@ -1,39 +1,39 @@
...
-  onmouseup={{handleMouseUpByBlurring}}
->
-  {{polaris-icon source="arrow-right"}}
-</button>
+  <button
+    data-test-next-btn="true"
+    class="Polaris-Pagination__Button"
+    aria-label="Next"
+    disabled={{this.isNextDisabled}}
+    onclick={{fn (or @onNext this.onNext)}}
```
vladucu: can we use modifiers here too? Ideally, we'd want to be consistent and use them everywhere instead of the old event bindings

addon/templates/components/polaris-pagination.hbs
```diff
@@ -1,39 +1,39 @@
...
+  ...attributes
 >
-  {{polaris-icon source="arrow-left"}}
-</button>
+  <button
+    data-test-prev-btn="true"
+    class="Polaris-Pagination__Button"
+    aria-label="Previous"
+    disabled={{this.isPreviousDisabled}}
+    {{on "click" (fn (or @onPrevious this.onPrevious))}}
```
vladucu: nit-pick: I think we can get away without `fn` helper usage....you'd generally need it in case you want to curry any arguments to the action

tomnez: Yeah it's not needed, I remember we used to explicitly use the `action` helper sometimes even when it wasn't required so I wasn't sure if we wanted to follow that practice with `fn` or not. I can stop doing this if we don't think it's necessary anymore.

vladucu: yeah, we won't need it anymore 🙌

PR #504, opened 2020-02-14
Modernize polaris-radio-button component

tomnez:
[DEV-76](https://smileio.atlassian.net/browse/DEV-76)

Updates `PolarisRadioButton` to tagless component, ES6 classes, and angle bracket syntax.

Dummy app:

radio

### DIFF
--- a/addon/templates/components/polaris-radio-button.hbs
+++ b/addon/templates/components/polaris-radio-button.hbs
@@ -16,7 +16,7 @@
       checked={{@checked}}
       disabled={{@disabled}}
       aria-describedby={{this.describedBy}}
-      onchange={{action (or @onChange this.onChange) value="target.value"}}
+      {{on "change" (or @onChange this.onChange)}}
       {{on "focus" (or @onFocus this.onFocus)}}
       {{on "blur" (or @onBlur this.onBlur)}}
     >
--- a/addon/templates/components/polaris-radio-button.hbs
+++ b/addon/templates/components/polaris-radio-button.hbs
@@ -17,8 +17,8 @@
       disabled={{@disabled}}
       aria-describedby={{this.describedBy}}
       onchange={{action (or @onChange this.onChange) value="target.value"}}
-      onfocus={{fn (or @onFocus this.onFocus)}}
-      onblur={{fn (or @onBlur this.onBlur)}}
+      {{on "focus" (or @onFocus this.onFocus)}}
+      {{on "blur" (or @onBlur this.onBlur)}}
     >
 
     <span data-test-radio-button-backdrop="true" class="Polaris-RadioButton__Backdrop"></span>
--- a/addon/templates/components/polaris-radio-button.hbs
+++ b/addon/templates/components/polaris-radio-button.hbs
@@ -1,27 +1,27 @@
-{{#polaris-choice
-  inputId=_id
-  label=label
-  labelHidden=labelHidden
-  disabled=disabled
-  helpText=helpText
-}}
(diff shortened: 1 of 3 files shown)


addon/templates/components/polaris-radio-button.hbs
```diff
@@ -1,27 +1,27 @@
...
-      onchange={{action onChange value="target.value"}}
-      onfocus={{action onFocus}}
-      onblur={{action onBlur}}
+      id={{this._id}}
+      name={{@name}}
+      value={{@value}}
+      checked={{@checked}}
+      disabled={{@disabled}}
+      aria-describedby={{this.describedBy}}
+      onchange={{action (or @onChange this.onChange) value="target.value"}}
```
tomnez: Left the `action` helper here because the `fn` wasn't working as expected when passing `value="target.value"` 🤷‍♂

vladucu: @tomnez can you try 
`<input ...... {{on "change" (or @onChange this.onChange)}} ....>`

addon/templates/components/polaris-radio-button.hbs
```diff
@@ -1,27 +1,27 @@
...
+      onfocus={{fn (or @onFocus this.onFocus)}}
```
vladucu: same...let's use the `on` modifier for these last 2

addon/templates/components/polaris-radio-button.hbs
tomnez: This still passes the `event` as the first (and only) argument now rather than the expected value 🤷‍♂

vladucu: not really a problem, we can just handle in the local action right? as the example [here](https://guides.emberjs.com/release/upgrading/current-edition/action-on-and-fn/)

Screen Shot 2020-02-17 at 4 09 59 PM

tomnez: Yeah we can do that 👍

addon/templates/components/polaris-radio-button.hbs
```diff
@@ -1,27 +1,27 @@
...
-      onchange={{action onChange value="target.value"}}
-      onfocus={{action onFocus}}
-      onblur={{action onBlur}}
+      id={{this._id}}
+      name={{@name}}
+      value={{@value}}
+      checked={{@checked}}
+      disabled={{@disabled}}
+      aria-describedby={{this.describedBy}}
+      {{on "change" (or @onChange this.onChange)}}
```
vladucu: I think we still want to pass the value to the `onChange` action....assuming we were correct and that's what the react implementation does
PS: we shouldn't modify tests to match our implementation.....but make sure tests are testing correct behavior and we pass 😬

vladucu: check the [example](https://github.com/smile-io/ember-polaris/pull/504#discussion_r380267497) I've posted earlier....that should be in the component not in the tests
let me know if this doesn't makes sense pls

tomnez: Ah so you mean we want a middle-man method (something like a private `handleChange` action) where we grab the value and _then_ call `onChange`?

Originally when you said 

> we can just handle in the local action

I thought you meant that the passed-in `onChange` would be responsible for this. Sorry about the confusion 😬

vladucu: yeah, like `handleChange`
moving this to the caller `onChange` would mean we break the API of this component...which we don't want

PR #511, opened 2020-02-18
Modernize polaris-popover components

tomnez:
[DEV-74](https://smileio.atlassian.net/secure/RapidBoard.jspa?rapidView=6&modal=detail&selectedIssue=DEV-74)

Updates `PolarisPopover` and subcomponent to tagless components, ES6 classes, and angle bracket syntax.

Dummy app:

popover

### DIFF
--- a/addon/components/polaris-popover.js
+++ b/addon/components/polaris-popover.js
@@ -121,8 +121,22 @@ export default class PolarisPopover extends Component {
   /**
    * @private
    */
+  triggerStyle = htmlSafe(`
+    display: inline-block;
+    overflow: inherit;
+    border: none;
+  `);
+
+  /**
+   * @private
+   */
   @computed('preferredPosition')
   get verticalPosition() {
+    // If `preferredPosition` is set to `mostSpace`, the value
+    // will be calculated and set when the user opens the popover.
+    // The only allowed values are 'above' and 'below', so we
+    // return null for anything other than those values and let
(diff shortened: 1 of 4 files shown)


addon/components/polaris-popover.js
```diff
@@ -117,46 +116,41 @@ export default Component.extend({
...
 
   /**
    * @private
    */
-  verticalPosition: computed('preferredPosition', {
-    // If `preferredPosition` is set to `mostSpace`, the value
-    // will be calculated and set when the user opens the popover.
-    // The only allowed values are 'above' and 'below', so we
-    // return null for anything other than those values and let
-    // ember-basic-dropdown use its default value.
```
andrewpye: Curious if there was a reason for dropping this comment

addon/components/polaris-popover.js
```diff
@@ -117,46 +116,41 @@ export default Component.extend({
...
+
+    return null;
+  }
 
   /**
    * @private
    */
-  triggerStyle: computed(function() {
+  @computed
+  get triggerStyle() {
```
andrewpye: Can this be assigned as a regular property now?

addon/components/polaris-popover.js
tomnez: I didn't realize it got dropped 🤔I want to blame a codemod but that doesn't sound right.. I'll add it back.

PR #602, opened 2020-06-19
Modernize text-container component

vladucu:


addon/components/polaris-text-container.js
```diff
@@ -1,53 +1,43 @@
...
 
-  /**
-   * @private
-   */
-  spacingClass: computed('spacing', function() {
-    let spacing = this.get('spacing');
-    if (allowedSpacings.indexOf(spacing) > -1) {
-      return `Polaris-TextContainer--spacing${classify(spacing)}`;
+  @(computed('spacing').readOnly())
+  get spacingClass() {
```
sivakumar-kailasam, line 36: we don't need to mark this as `readOnly` since we've only defined the getter.

vladucu: 💯

PR #611, opened 2020-06-22
Modernize spinner component

vladucu:


vladucu:
> There's a couple CP's that reuse the same value rather than setting a var (this.color and this.size) but I'm fine with leaving it if no one else minds it 🤷

those 👁️ 
yeah, I think that should be fine... we shouldn't always just reach for destructuring IMO

tomnez:
> yeah, I think that should be fine... we shouldn't always just reach for destructuring IMO

kk, I'll ignore if it pops up in other PR's then 👍

PR #614, opened 2020-06-22
Modernize empty-state component

vladucu:


addon/components/polaris-empty-state.js
```diff
@@ -55,29 +49,24 @@ export default Component.extend({
...
 
   /**
    * Primary action for empty state
    *
-   * @property action
    * @type {Object}
    * @default null
    */
-  action: null,
+  action = null;
```
tomnez, line 63: Does this break in our version of Ember here? I thought we had to update the use of attributes named `action` at some point in our Ember upgrades but maybe I'm remembering wrong.

addon/templates/components/polaris-empty-state/details.hbs
```diff
@@ -1,37 +1,42 @@
...
-      {{/if}}
-    </PolarisStack>
-  </div>
-{{/if}}
+        {{#if @secondaryAction}}
+          <stack.item>
+            <PolarisButton
+              @plain={{true}}
+              @text={{@secondaryAction.text}}
+              @onClick={{action @secondaryAction.onAction}}
```
tomnez, line 35: `fn` helper?

vladucu: 🙌

addon/components/polaris-empty-state.js
vladucu: good catch.... it seems to work, but we should probably do this just to avoid confusions
have this noted to recheck once all components are done

PR #615, opened 2020-06-23
Modernize connected component

vladucu:


addon/templates/components/polaris-connected/item.hbs
```diff
@@ -1 +1,9 @@
-{{yield}}
+{{!-- template-lint-disable no-invalid-interactive --}}
+<div
+  class={{this.classes}}
+  ...attributes
+  {{on "focus" this.handleFocus}}
+  {{on "blur" this.handleBlur}}
```
andrewpye, line 6: Do we need to prevent bubbling on these events? 🤔

vladucu: I don't think anything changes from what we had before with this transition
but we'll definitely find out once we start testing

PR #616, opened 2020-06-23
Modernize empty-search-result component

vladucu:


addon/components/polaris-empty-search-result.js
```diff
@@ -1,32 +1,28 @@
 import Component from '@ember/component';
+import { tagName, layout as templateLayout } from '@ember-decorators/component';
 import layout from '../templates/components/polaris-empty-search-result';
```
andrewpye, line 3: In #615 we've imported the template as `template` instead of `layout` so that we don't have to rename the `layout` decorator to `templateLayout` - I think that's cleaner than what's here, what are your thoughts about updating this for consistency/cleanness?

vladucu: yeah, I've been doing that manually initially
but the codemod uses this, so just went with it and will probably just update few occurrences I did manually to this for consistency

PR #618, opened 2020-06-23
Modernize description-list component

vladucu:


addon/templates/components/polaris-description-list.hbs
```diff
@@ -1,16 +1,18 @@
...
-    {{else}}
-      {{render-content item.description}}
-    {{/if}}
-  </dd>
-{{/each}}
+<dl class="Polaris-DescriptionList">
+  {{#each @items as |item|}}
+    <dt class="Polaris-DescriptionList__Term">
+      {{#if item.termComponent}}
+        {{component item.termComponent}}
```
andrewpye, line 5: Since we've removed the deprecation warning for `termComponent`, do we want to drop support here? Same below re. `descriptionComponent` too.

vladucu: oh, for sure
I thought it's just some leftovers

PR #619, opened 2020-06-23
Modernize date-picker component

vladucu:
Had to skip some tests for the `ResourceList` which is using the date-picker. I'd like to get this in since it's the last component and as part of a cleanup re-circle back to theses tests.


vladucu:
> Whew these files give me flashbacks 😅 
lmao

> Looks good assuming it got a good test run.

welp, fun only now begins 😬

PR #620, opened 2020-06-26
Misc updates for v6

vladucu:
- adds few files missed for `text-field` component during modernization PR
- deprecates passing `@tagName` to some components and replaces it with `htmlTag`
- some cleanup: no `.get` where it's not needed
- fixes the version for new deprecations added - _this will be `v6` so we'll support those through to `v7`_
- [Breaking] removes long deprecated `labelComponent` for `choice` component

### DIFF
--- a/addon/templates/components/polaris-display-text.hbs
+++ b/addon/templates/components/polaris-display-text.hbs
@@ -1,4 +1,4 @@
-{{#let (element (or (or @htmlTag @tagName) "p")) as |DisplayText|}}
+{{#let (element (or @htmlTag @tagName "p")) as |DisplayText|}}
   <DisplayText
     class="Polaris-DisplayText {{this.sizeClassName}}"
     data-test-display-text
--- a/addon/templates/components/polaris-display-text.hbs
+++ b/addon/templates/components/polaris-display-text.hbs
@@ -1,4 +1,4 @@
-{{#let (element (or @htmlTag "p")) as |DisplayText|}}
+{{#let (element (or (or @htmlTag @tagName) "p")) as |DisplayText|}}
   <DisplayText
     class="Polaris-DisplayText {{this.sizeClassName}}"
     data-test-display-text
--- a/addon/templates/components/polaris-text-field.hbs
+++ b/addon/templates/components/polaris-text-field.hbs
@@ -1,121 +1,104 @@
-{{#polaris-labelled
-  dataTestLabelled=dataTestTextField
-  id=id
-  label=label
-  error=error
-  action=labelAction
-  labelHidden=labelHidden
-  helpText=helpText
-}}
-
-  {{#polaris-connected
-    left=connectedLeft
-    right=connectedRight
-  }}
+<PolarisLabelled
+  @dataTestLabelled={{this.dataTestTextField}}
+  @id={{this.id}}
+  @label={{@label}}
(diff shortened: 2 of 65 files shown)


addon/templates/components/polaris-text-field/resizer.hbs
```diff
@@ -5,7 +5,29 @@
...
+    {{this.finalContents}}
+  </div>
+
+  {{#if @minimumLines}}
+    <div
+      class="Polaris-TextField__DummyInput"
+      {{did-insert this.setMinimumLinesNode}}
+    >
+      {{this.contentsForMinimumLines}}
+    </div>
```
andrewpye, line 31: Have you checked that multiline text fields render/resize to fit their content correctly with these changes? There's a comment at the top of this file that says these two dummy inputs need to be on a single line to render the correct size. If that's no longer the case, can we remove the comment?

addon/templates/components/polaris-text-field/resizer.hbs
```diff
@@ -5,7 +5,29 @@
...
+  aria-hidden="true"
+  data-test-text-field-resizer
+  ...attributes
+  {{on "resize" this.handleHeightCheck}}
+  {{did-insert this.handleHeightCheck}}
+  {{did-update this.handleHeightCheck @contents @currentHeight @minimumLines}}
+>
+  <div
+    class="Polaris-TextField__DummyInput"
+    {{did-insert this.this.setContentNode}}
```
andrewpye, line 20: Extra `this.` in here 😬 

```suggestion
    {{did-insert this.setContentNode}}
```

addon/templates/components/polaris-display-text.hbs
```diff
@@ -1,4 +1,4 @@
-{{#let (element (or @htmlTag "p")) as |DisplayText|}}
+{{#let (element (or (or @htmlTag @tagName) "p")) as |DisplayText|}}
```
andrewpye, line 1: Can just be

```suggestion
{{#let (element (or @htmlTag @tagName "p")) as |DisplayText|}}
```

instead of nesting `or`s? Same applies in a few other places where you've added `htmlTag` support.

addon/templates/components/polaris-text-field.hbs
```diff
@@ -1,121 +1,104 @@
...
+          autocomplete={{this.autoCompleteInputs}}
+          max={{@max}}
+          min={{@min}}
+          step={{@step}}
+          minlength={{@minLength}}
+          maxlength={{@maxLength}}
+          spellcheck={{@spellCheck}}
+          value={{this.normalizedValue}}
+          class={{this.inputClassName}}
+          aria-multiline={{@multiline}}
```
andrewpye, line 47: Does this add `aria-multiline="false"` or anything similar if `@multiline={{true}}` is passed to the `PolarisTextField`? If so, can we check if that replicates the React implementation and if not, update it? 🙇

addon/templates/components/polaris-text-field/resizer.hbs
vladucu: checked now, fixed some issues 🙌

addon/templates/components/polaris-text-field.hbs
vladucu: made sure it matches react implementation

PR #630, opened 2020-06-29
Stop curring click event to action handlers

vladucu:
When we replaced usage of `mapEventToAction` utility for click handlers, we
introduced back currying of the click event to the action handlers. This fixes
it.


vladucu:
> Makes sense to me I think. One question that I had is whether we need to stop event propagation in our handleClick actions before invoking onClick etc.

Will check to make sure we match React implementation, just to be sure. Good shout!

PR #633, opened 2020-07-08
Fix PolarisButton click propagation

vladucu:
This fixes the issue we've been seeing in our testing with the DropZone
click handling not working correctly.


vladucu:
> Does this change how our button handles the click event compared to how the React implementation handles it?

This just keeps the same functionality as we had before, not sure about Polaris

> Also noticed that CI is erroring, is that something that'll be fixed in a separate PR?

seems unrelated to these tests

PR #634, opened 2020-07-08
Add deprecations for v6

vladucu:
We've previously made a breaking change for `PolarisBanner` where we renamed `action` argument `primaryAction`.

This adds backwards support for `action` argument too together with a deprecation. Same changes are included for all the other components that have been using `action` as an argument.

Also, stops using a mixin for adding a deprecation re `class` argument that was implicitly working previously with non-tagless components in Ember.


addon/components/polaris-avatar.js
```diff
@@ -25,11 +25,10 @@ const avatarImages = [
 
 const styleClasses = ['one', 'two', 'three', 'four', 'five', 'six'];
 
+@deprecateClassArgument
```
andrewpye, line 28: This is neat, but what do you think of having a generic `deprecateArgument` decorator instead so that we have something reusable going forward? Maybe gets a bit too clunky having to pass the various deprecation properties (message, version etc.)?

addon/components/polaris-empty-state/details.js
```diff
@@ -1,46 +1,7 @@
...
-
-  /**
-   * Secondary action for empty state
-   *
-   * @type {Object}
-   * @default null
-   */
-  secondaryAction = null;
-}
+export default class PolarisEmptyStateDetails extends Component {}
```
andrewpye, line 7: Do we still need this JS file? Also curious why the deprecation warning's been removed 🤔

addon/templates/components/polaris-empty-state/details.hbs
```diff
@@ -14,7 +14,7 @@
     </div>
   </PolarisTextContainer>
 
-  {{#if action}}
+  {{#if @action}}
```
andrewpye, line 17: Should this be renamed as well? 🤔

addon/components/polaris-empty-state/details.js
vladucu: Seems still needed in addons?!
This is just internal, so no need to repeat the deprecations and documenting properties since we do in the parent component. Ideally, this would be just a tagless component

addon/templates/components/polaris-empty-state/details.hbs
vladucu: yeap

addon/components/polaris-avatar.js
vladucu: yeah, I def like how simple the usage with this is
but something we should keep in mind and maybe do in the future 👍

vladucu:
> Nice cleanup @vladucu! I would have preferred to see the class deprecation changes done as a separate PR as they're unrelated to the scope suggested by this PR's title. Perhaps we could update the title instead so that it more accurately reflects what's happening in this PR?

agree and sorry.... just trying to pick my battles 😂 

> Wondering if it's worth introducing a decorator for the action deprecation - no strong views either way though.

I'd lean just leaving this as is, since it's just temporary code

PR #649, opened 2020-08-19
Pass rows attribute to did-update modifier in PolarisDataTable

tomnez:
Cell heights aren't getting recalculated once the `cellText` (which comes from `@rows` data) gets rendered inside the cell, so its resulting in broken cells where they're are all the initial default height regardless of how much content is rendered inside of them.


andrewpye (approved):
If this works... :shipit: 🚀

PR #654, opened 2020-09-23
Use ember-cli-htmlbars in tests

vladucu:
Starts using `ember-cli-htmlbars` instead of the deprecated
`htmlbars-inline-precompile`.  Changes are the result of running the
[codemod](https://github.com/ember-codemods/ember-cli-htmlbars-inline-precompile-codemod)


andrewpye (approved):
Good with me if tests pass! 🚀

PR #655, opened 2020-09-23
Upgrade ember-keyboard to v6

vladucu:
This fixes failing tests on `ember-beta`/`ember-canary`. This is a major
upgrade for `ember-keyboard` because it drops support for Ember pre `3.8`, but we are compatible with
`3.12+`, so not a major for us.

There are still some deprecations, we'll fix these later on.


andrewpye (approved):
Niiiice! 🚀

PR #656, opened 2020-09-24
Polaris styles build pipeline updates

vladucu:
Adds updated Polaris styles build pipeline to work with polaris-react `v5`. For more see [here](https://github.com/Shopify/polaris-react/blob/master/documentation/guides/migrating-from-v4-to-v5.md#build-output-changes---general-consumers)

In summary, latest Polaris v5 changes include:
- main styles are no longer provided in SASS, but plain css
- there's still an optional SASS file with mixins and functions that we import _(ideally we'll make this import and having the consuming apps use SASS preprocessor optional altogether)_

**NOTE** This is still experimental and might be tweaked down the road.


index.js
```diff
@@ -4,6 +4,7 @@ const path = require('path');
 const resolve = require('resolve');
 const Funnel = require('broccoli-funnel');
 const MergeTrees = require('broccoli-merge-trees');
+const BroccoliDebug = require('broccoli-debug');
```
andrewpye, line 7: Is this something you added for testing this out, to be removed later? Or to keep in to help consumers?

andrewpye (approved):
Can't say I understand a huge amount of what the changes in the build are doing, but I know you wouldn't have put this up without verifying it works 😝

index.js
vladucu: I was thinking to keep it, there's no difference for consuming apps, other addons do this too.

It just helps debugging easily the build tree. For example running `BROCCOLI_DEBUG=ember-polaris:* ember b` will log the trees at each step and put it in a `DEBUG` folder.

PR #657, opened 2020-10-01
Update icons handling

vladucu:
1. Updates `PolarisIcon` to be compatible with latest `polaris-react#v5`
version
2. Updates build so that consuming apps no longer need to manually include
Polaris icons.  New icons can still be added through `ember-svg-jar`, or they
can be passed directly as SVG's now.

**NOTEs:** 
1. Consuming apps should continue to work correctly with these changes,
  assuming previous Polaris icons have been set up correctly as documented
  through `ember-svg-jar`.
2. Latest `<PolarisIcon>` allows passing an SVG as a string directly that will be rendered as an untrusted SVG using an `img` tag. [Example](https://github.com/smile-io/ember-polaris/pull/657/files#diff-c43f6653fd914eb22dbd7c06a0b4a143R78)


addon/components/polaris-icon.hbs
```diff
@@ -0,0 +1,25 @@
+<span
```
andrewpye, line 1: Out of curiosity, why isn't this in `addon/templates/components`? 🤔

docs/icon.md
```diff
@@ -2,22 +2,40 @@
...
 ```hbs
-{{polaris-icon source="add" color="darkTeal" backdrop=true}}
+<PolarisIcon @source="AddMinor" @color="tealDark" @backdrop={{true}} />
+```
+
+
+### Rendering custom icons
+
+By default, `PolarisIcon` will render SVG's passed directly as strings to `@source` as external images.
+If you want to render custom icons similarly as main Polaris icons are rendered, you can configure `ember-svg-jar` in your app instead.
```
andrewpye, line 25: Possible slight tweak for grammar:

```suggestion
If you want to render custom icons in the same way that the built-in Polaris icons are rendered, you can configure `ember-svg-jar` in your app instead.
```

docs/migrating-from-v6-to-v7.md
```diff
@@ -0,0 +1,21 @@
+# Migrating from v6 to v7
+
+## Component
+
+### PolarisIcon
+
+<!-- TODO #polaris-v5 consider adding a codemod to migrate consuming apps -->
+Working with Polaris icons has been simplified and icons are now automatically included in the consuming app.
+There are now more Polaris icons and are PascalCased to be consistent with polaris-react.
```
andrewpye, line 9:
```suggestion
There are now more Polaris icons and they are PascalCased to be consistent with `polaris-react`.
```

docs/migrating-from-v6-to-v7.md
```diff
@@ -0,0 +1,21 @@
...
+```hbs
+<!-- Old -->
+<PolarisIcon @source="notes" />
+
+<!-- New -->
+<PolarisIcon @source="NotesMajor" />
+```
+
+**NOTE**
+1. `PolarisIcon` component leverages [ember-svg-jar](https://github.com/ivanvotti/ember-svg-jar) under the hood, which means consuming apps can add other SVG assets to svg-jar which should be usable directly with this component.
```
andrewpye, line 20:
```suggestion
1. The `PolarisIcon` component leverages [`ember-svg-jar`](https://github.com/ivanvotti/ember-svg-jar) under the hood, which means consuming apps can add other SVG assets to `svg-jar` which should be usable directly with this component.
```

tests/integration/components/polaris-action-list-test.js
```diff
@@ -3,7 +3,7 @@ import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 import { findAll, click, render } from '@ember/test-helpers';
 import buildNestedSelector from '../../helpers/build-nested-selector';
-import MockSvgJarComponent from '../../mocks/components/svg-jar';
```
andrewpye, line 6: Can we remove the mock SVG jar component now?

andrewpye (commented):
This is great stuff @vladucu! 🙌 🔥 

Left a couple of suggestions for grammar tweaks in the docs, and I was also wondering if removing the SVG handling mixin for removing fills etc. could impact any other components (either in `ember-polaris` or in downstream usages) 🤔

addon/components/polaris-icon.hbs
vladucu: This co-locates the template and component file. Change done through [codemod](https://github.com/ember-codemods/ember-component-template-colocation-migrator) & [RFC here](https://github.com/emberjs/rfcs/pull/481).
This will be needed for all components in order to support Octane-only consuming apps.

Btw, sorry, wanted to, but completely forgot to mention this in the description.

PS: These can also be [nested](https://github.com/ember-codemods/ember-component-template-colocation-migrator#nested-component-structure), but for now went with the default mode. We can change later probably if we feel it's better for us

tests/integration/components/polaris-action-list-test.js
vladucu: ayyy!

andrewpye (approved):
🚀

PR #659, opened 2020-10-06
Drop ember-invoke-action dependency

vladucu:


addon/components/polaris-action-list/section.js
```diff
@@ -40,11 +39,15 @@ export default class PolarisActionListSection extends Component {
...
     event.stopPropagation();
 
-    invokeAction(item, 'onAction');
-    invokeAction(this, 'onActionAnyItem');
+    if (item.onAction) {
+      item.onAction();
+    }
+    if (this.onActionAnyItem) {
+      this.onActionAnyItem();
+    }
```
andrewpye, line 51: How do we feel about using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to clean this up?

```suggestion
    item.onAction?.();
    this.onActionAnyItem?.();
```

andrewpye (approved):
Niiiiice 😁

PR #664, opened 2020-10-12
Make <VisuallyHidden> a Glimmer component

vladucu:
- makes `<PolarisVisuallyHidden>` a Glimmer component and co-locates the template
- no other changes required for comaptibility with `v5`


addon/components/polaris-visually-hidden.js
```diff
@@ -1,15 +1,11 @@
...
-import { tagName, layout } from '@ember-decorators/component';
-import template from '../templates/components/polaris-visually-hidden';
-import deprecateClassArgument from '../utils/deprecate-class-argument';
+import Component from '@glimmer/component';
 
 /**
  * Polaris Visually hidden component.
  * See https://polaris.shopify.com/components/titles-and-text/visually-hidden
+ *
+ * @version 5.6.0
```
andrewpye, line 7: Nice idea! 👍

addon/components/polaris-visually-hidden.js
```diff
@@ -18,9 +14,8 @@ export default class PolarisVisuallyHidden extends Component {
    * in which case the block content will be used
    * instead of `text`
    *
-   * @type {string}
-   * @default null
+   * @type {String}
    * @public
    */
-  text = null;
+  text;
```
andrewpye, line 20: What are your thoughts about needing these prop declarations, especially when we don't need default values (like here)?

andrewpye (approved):
:shipit:

addon/components/polaris-visually-hidden.js
vladucu: I'm flexible, left it for now
Wouldn't mind cleaning them up, but I thought that since it's already there, no harm

PR #665, opened 2020-10-12
Update <PolarisActionList>

vladucu:
- updates `PolarisActionList` to match `polaris-react#v5.6.0`
- co-locates templates & Refactors into Glimmer component
- adds support for some previously unsupported item attributes, ex: `badge`, `url`, `ellipsis`


addon/components/polaris-action-list.js
```diff
@@ -1,74 +1,50 @@
...
+export default class PolarisActionListComponent extends Component {
   /**
    * Collection of actions for list
-   *
    * @type {Array}
-   * @default []
-   * @public
    */
-  items = [];
+  items;
```
andrewpye, line 15: Same comment for these as [here](https://github.com/smile-io/ember-polaris/pull/664/files#r503747187) re. the value of declaring these if we're not assigning default values 🤔

addon/components/polaris-action-list/item.js
```diff
@@ -113,13 +115,12 @@ export default class PolarisActionListItem extends Component {
...
 
     return cssClasses.join(' ');
   }
 
-  @computed('image')
   get imageBackgroundStyle() {
-    let { image: url } = this;
-    return url ? htmlSafe(`background-image: url(${url})`) : '';
+    return htmlSafe(`background-image: url(${this.args.image})`);
   }
```
andrewpye, line 125: Not sure how this works with glimmer, will this update automatically when `args.image` changes (i.e. are properties under `args` automatically `@tracked`)?

andrewpye (approved):
Works for me!

addon/components/polaris-action-list/item.js
vladucu: yes, arguments are auto-tracked 🕺

PR #669, opened 2020-10-28
Fix banner primary action

vladucu:
Fixes triggering the Banner's main action


abd00s (approved):
I'm out of my element here, but looks good!

PR #671, opened 2020-11-10
Update PolarisCaption component

vladucu:
- co-locates template
- misc 🧹


andrewpye (approved):
🚀

PR #672, opened 2020-11-10
Update CalloutCard component

vladucu:
See diff [here](https://github.com/Shopify/polaris-react/compare/v3.10.0...v5.5.0#diff-2fcc58c07de1af4a1f3c5d190625ec588a34d6d7bf79af26fc731a69b9c428a6R1), only change is the icon used for the dismiss button.

Also:
- co-locates template
- misc 🧹


addon/components/polaris-callout-card.hbs
```diff
@@ -55,12 +57,13 @@
...
                 </PolarisButtonGroup>
               {{else}}
-                <PrimaryButton />
+                <PrimaryButton data-test-callout-card-primary-button />
               {{/if}}
             {{/let}}
           </div>
         </div>
 
+        {{!-- TODO use <PolarisImage> once implemented  --}}
```
andrewpye, line 66: We might actually be able to implement this now that we have splattributes available to us 🤔

docs/callout-card.md
```diff
@@ -2,37 +2,37 @@
...
-  title="New feature"
-  text="This new feature is awesome!"
-  primaryAction=(hash
+<PolarisCalloutCard
+  @title="New feature"
+  @text="This new feature is awesome!"
+  @primaryAction=(hash
     text="Take a look"
-    onAction=(action "showNewFeature")
+    onAction=(fn this.handleAction)
```
andrewpye, line 17: Any reason for changing this name? It doesn't associate so clearly with the example card content like this 🤔

andrewpye (approved):
Good work @vladucu!

docs/callout-card.md
vladucu: haha, no
will restore

PR #675, opened 2020-11-11
Update polaris breadcrumbs

tomnez:
v3.10 - v5.5 diff [here](https://github.com/Shopify/polaris-react/compare/v3.10.0...v5.5.0)

- co-locates template
- misc 🧹


addon/components/polaris-breadcrumbs.js
```diff
@@ -16,7 +10,7 @@ export default class PolarisBreadcrumbs extends Component {
    * @default []
    * @public
    */
-  breadcrumbs = [];
```
vladucu, line 19: wouldn't removing this default blow up if breadcrumbs is undefined and we assume it's an array?

addon/components/polaris-breadcrumbs.hbs
```diff
@@ -0,0 +1,41 @@
...
+    <PolarisUnstyledLink
+      class="Polaris-Breadcrumbs__Breadcrumb"
+      @url={{this.breadcrumb.url}}
+      @ariaLabel={{this.breadcrumb.accessibilityLabel}}
+      @onMouseUp={{this.handleMouseUpByBlurring}}
+    >
+      <span class="Polaris-Breadcrumbs__ContentWrapper">
+        <span data-test-breadcrumb-icon class="Polaris-Breadcrumbs__Icon">
+          {{!-- TODO #polaris-v5-newDesignLanguage add newDesignLanguage icon, styles, and visually-hidden wrapper --}}
+          <PolarisIcon @source="chevron-left" />
```
vladucu, line 16:
```suggestion
          <PolarisIcon @source="ChevronLeftMinor" />
```

addon/components/polaris-breadcrumbs.hbs
```diff
@@ -0,0 +1,41 @@
...
+    <button
+      class="Polaris-Breadcrumbs__Breadcrumb"
+      type="button"
+      aria-label={{this.breadcrumb.accessibilityLabel}}
+      {{on "click" this.handleClick}}
+      {{on "mouseup" this.handleMouseUpByBlurring}}
+    >
+      <span class="Polaris-Breadcrumbs__ContentWrapper">
+        <span data-test-breadcrumb-icon class="Polaris-Breadcrumbs__Icon">
+          <PolarisIcon @source="chevron-left" />
```
vladucu, line 33:
```suggestion
          <PolarisIcon @source="ChevronLeftMinor" />
```

addon/components/polaris-breadcrumbs.hbs
```diff
@@ -0,0 +1,41 @@
+<nav
```
vladucu, line 1: if there is no `breadcrumb`, we shouldn't be rendering anything. Can we do that here like [original version](https://github.com/Shopify/polaris-react/blob/master/src/components/Breadcrumbs/Breadcrumbs.tsx#L27) pls

vladucu (changes requested):
Minor comments and we need to fix tests

addon/components/polaris-breadcrumbs.js
tomnez: Not in this case since we only use it in `get breadcrumb()` which won't run unless `breadcrumbs` changes (a.k.a. gets passed in as a value). But to your point maybe we should keep it in case things change in the future?

Should we create a getter with `this.args.breadcrumbs || []`, or just do a check inside `get breadcrumb()` since that's the only spot it's used? Any strong opinion here? @vladucu

addon/components/polaris-breadcrumbs.hbs
```diff
@@ -0,0 +1,41 @@
...
+  {{#if this.breadcrumb.url}}
+    <PolarisUnstyledLink
+      class="Polaris-Breadcrumbs__Breadcrumb"
+      @url={{this.breadcrumb.url}}
+      @ariaLabel={{this.breadcrumb.accessibilityLabel}}
+      @onMouseUp={{this.handleMouseUpByBlurring}}
+    >
+      <span class="Polaris-Breadcrumbs__ContentWrapper">
+        <span data-test-breadcrumb-icon class="Polaris-Breadcrumbs__Icon">
+          {{!-- TODO #polaris-v5-newDesignLanguage add newDesignLanguage icon, styles, and visually-hidden wrapper --}}
```
vladucu, line 15: @tomnez I've merged #678 which adds support for this new design language option.
Let's update this and remove the TODO before considering this done

addon/components/polaris-breadcrumbs.js
```diff
@@ -1,22 +1,19 @@
...
+
   /**
    * Collection of breadcrumbs
    *
    * @type {Array}
    * @default []
    * @public
    */
-  breadcrumbs = [];
+  breadcrumbs; // TODO add @arg breadcrumbs = [] when decorator is merged
```
tomnez, line 16: Will get this done before merging once we get my PolarisBadge PR merged.

vladucu (commented):
@tomnez there are some merge conflicts here, let's get those addressed before another review pls

vladucu (changes requested):
There are some failing specs related to changes here.

Note that [this](https://github.com/smile-io/ember-polaris/pull/677) was merged which is a component used by breadcrumbs.

vladucu (approved):
There's a pending TODO left, but otherwise this looks really nice.
Thanks for the updates, nicely done! 🚀

PR #676, opened 2020-11-12
Update PolarisBadge

tomnez:
v3.10 - v5.5 diff [here](https://github.com/Shopify/polaris-react/compare/v3.10.0...v5.5.0#diff-8d9d91aab0164a19825f5b7b18084a0ce9d38b32462d3486ca827ebe35038a40)

- co-locates template
- misc 🧹


addon/components/polaris-badge.js
```diff
@@ -1,11 +1,8 @@
-import Component from '@ember/component';
+import Component from '@glimmer/component';
 import { computed } from '@ember/object';
```
vladucu, line 2: this `computed` leftover is currently breaking tests

addon/components/polaris-badge.js
```diff
@@ -65,66 +60,61 @@ export default class PolarisBadge extends Component {
...
   /**
    * Medium or small size. Use `small` only in the main navigation of an app frame..
    *
    * @type {String}
-   * @default 'medium'
+   * @default null
    * @public
    */
-  size = DEFAULT_SIZE;
+  size;
```
vladucu, line 72: any reason we removed the default? it seems it's still the case [here](https://github.com/Shopify/polaris-react/blob/master/src/components/Badge/Badge.tsx#L49)

addon/components/polaris-badge.js
```diff
@@ -65,66 +60,61 @@ export default class PolarisBadge extends Component {
...
     }
 
     return PROGRESS_LABELS[progress];
   }
 
-  @computed('progress')
   get progressClass() {
-    const { progress } = this;
+    const { progress } = this.args;
     if (isBlank(progress) || progress === 'default') {
```
vladucu, line 91: same here with `default` (and below for `status` too)

addon/components/polaris-badge.js
```diff
@@ -65,66 +60,61 @@ export default class PolarisBadge extends Component {
...
   @notEmpty('progress')
   hasProgress;
 
   @notEmpty('status')
   hasStatus;
 
-  @computed('progress')
   get progressDescription() {
-    const { progress } = this;
+    const { progress } = this.args;
```
vladucu, line 81: I know this is unchanged, wondering why we have this `default` comparison here.

If it turns out it's not needed, we can likely simplify this a bit, example
```js
  get progressDescription() {
    return this.args.progress ? PROGRESS_LABELS[this.args.progress] : null;
  }
```

vladucu (commented):
Looks pretty good to me, minor questions

addon/components/polaris-badge.hbs
```diff
@@ -1,6 +1,7 @@
+{{!-- TODO #polaris-v5-newDesignLanguage add WithinFilterContext utility and style --}}
```
vladucu, line 1: @tomnez I've merged #678 which adds support for this new design language option.
Let's update this and remove the TODO before considering this done

addon/components/polaris-badge.js
```diff
@@ -65,44 +60,43 @@ export default class PolarisBadge extends Component {
...
    */
-  progress = null;
+  progress;
 
   /**
    * Medium or small size. Use `small` only in the main navigation of an app frame..
    *
    * @type {String}
-   * @default 'medium'
+   * @default null
```
vladucu, line 69:
```suggestion
   * @default 'medium'
```

addon/components/polaris-badge.js
vladucu: Nice job with `argWithDefault` decorator 👏 

I've seen this other format in [ember-bootstrap](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/bs-button.js#L186) which I've also found nice,  closer to plain JS and love the short name. Interested to hear what you think cc @andrewpye 
ex:
```js
@arg
size = DEFAULT_SIZE
```

addon/components/polaris-badge.js
```diff
@@ -112,23 +106,17 @@ export default class PolarisBadge extends Component {
...
-      return null;
-    }
+    return this.args.status
+      ? `Polaris-Badge--status${classify(this.args.status)}`
+      : null;
+  }
 
-    return `Polaris-Badge--status${classify(status)}`;
+  get withinFilterClass() {
+    return this.withinFilterContext ? 'Polaris-Badge--withinFilter' : null;
```
vladucu, line 120: should this be `this.args.withinFilterContext`?

vladucu (commented):
Looks pretty good to me overall!

addon/components/polaris-badge.js
andrewpye: I also like the short `@arg` syntax for its brevity and for keeping the default value "inline" with the prop declaration (sorry @tomnez 😝)

addon/components/polaris-badge.js
```diff
@@ -65,44 +60,43 @@ export default class PolarisBadge extends Component {
    * @default null
    * @public
    */
-  progress = null;
+  progress;
 
   /**
    * Medium or small size. Use `small` only in the main navigation of an app frame..
```
andrewpye, line 66: Well this bugs me (so this is probably where @tomnez adds 7 more `.`s at the end 😂)

```suggestion
   * Medium or small size. Use `small` only in the main navigation of an app frame.
```

andrewpye (commented):
Good work @tomnez! 💪

addon/components/polaris-badge.js
```diff
@@ -65,44 +60,43 @@ export default class PolarisBadge extends Component {
...
 
-  @computed('progress')
   get progressDescription() {
-    const { progress } = this;
-    if (isBlank(progress) || progress === 'default') {
-      return null;
-    }
-
-    return PROGRESS_LABELS[progress];
+    return this.args.progress ? PROGRESS_LABELS[this.args.progress] : null;
```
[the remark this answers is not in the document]
vladucu: Yeah, me and @tomnez have been chatting about this and we couldn't figure out where this `default` value was coming from, since the official Polaris implementation one doesn't really has anything like this, so we just agreed to drop it.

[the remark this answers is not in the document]
andrewpye: Coolio. I have a suspicion that I added it in our initial implementation back when I was building components based on the HTML generated in the React examples rather than porting the React code itself 😅 Thanks for cleaning it up! 🙌

addon/components/polaris-badge.js
tomnez: This attr uses our `@argWithDefault` decorator with a default of `false`, so we don't need to go via `args` here since the decorator does that for us.

addon/components/polaris-badge.js
tomnez: Ha I'm good with this too. I liked the `WithDefault` part of the name though, I know `@arg` is nice and short but it isn't really descriptive as to why it's being used in some places but not others for named arguments 🤷

andrewpye (approved):
It looks like we need to update `ember-basic-dropdown` to fix tests on Canary, but approving anyway since that should be a separate PR.

addon/components/polaris-badge.js
vladucu: oh, yes, pls ignore my comment 🤦

vladucu (approved):
@tomnez let's just setup to allow the canary to fail for now in this PR, so it's not a blocker here and we'll get to that upgrade too soon

andrewpye:
> It looks like we need to update `ember-basic-dropdown` to fix tests on Canary, but approving anyway since that should be a separate PR.

@tomnez you should now be able to merge master up to handle this 👍

vladucu:
> @tomnez you should now be able to merge master up to handle this 👍

let's merge `master` in `v7` and have `v7` branch merged in here to fix this pls

tomnez:
Ok, merged `master` into `v7` and pushed it up. The updated `v7` is now pulled into this branch.

PR #677, opened 2020-11-12
Update private UnstyledLink component

vladucu:
- co-locates templates
- 💣 removes deprecation & backwards compatibility (`dataTestId`, `dataPolarisUnstyled`, `id`)
- 💣 removes a bunch of arguments (`ariaLabel`, `ariaDescribedBy`, `onFocus`, `onBlur`, `onMouseUp`) which should be just passed as HTML attributes when used in angle-bracket syntax


tests/integration/helpers/no-op-test.js
```diff
@@ -0,0 +1,15 @@
...
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Helper | no-op', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('It successfully renders and does nothing when clicked', async function (assert) {
+    assert.expect(0);
+
+    await render(hbs`<PolarisButton @onClick={{fn (no-op)}}></PolarisButton> `);
+    await click('button');
```
andrewpye, line 13: Perhaps I'm missing something - how does this this test that the helper no-ops? 🤔

addon/helpers/no-op.js
```diff
@@ -0,0 +1,7 @@
+import { helper } from '@ember/component/helper';
+
+export function noOp() {
+  return () => {};
+}
+
+export default helper(noOp);
```
andrewpye, line 7: Is this helper a leftover or from another branch? It doesn't seem to be used in this PR 👀

vladucu: it's from the first iteration of changes, but eventually removed more arguments (ex: `onFocus`, `onBlur`, etc), so was no longer needed.
Kept it as it might be useful for other updates, but don't mind removing it either

tests/integration/helpers/no-op-test.js
vladucu: welp, we're just testing that there's no assertion, not sure what more can be done to test this

addon/helpers/no-op.js
andrewpye: I'd say let's add it when we need it, otherwise it's outside the scope of this PR.

vladucu: thanks!
generally, agree with that feedback, in this case I think it's just more work to remove now and add back later, since it's so small and unimportant piece, but will happily update if you think that's better

regardless, do you think this ^ should be a blocker here?

andrewpye: OK, if we're definitely going to use it elsewhere and you prefer to keep it 🤷‍♂️

tests/integration/helpers/no-op-test.js
andrewpye: >not sure what more can be done to test this

Me neither 😝 But for there to be an assertion, that'd mean that `no-op` itself would have to contain an `assert`, which wouldn't make any sense anyway. Perhaps we can just make sure it doesn't throw any exception when clicked instead?

andrewpye (approved):

tests/integration/helpers/no-op-test.js
vladucu: > Perhaps we can just make sure it doesn't throw any exception when clicked instead?

yeah, exactly. I think that's already implied, since tests pass. If there would be an exception, tests would fail.

andrewpye: >I think that's already implied, since tests pass. If there would be an exception, tests would fail.

Not quite though right? Wouldn't the tests _error_ instead of fail? Isn't it better to be explicit about what we're testing, instead of saying "there aren't any assertions in this test"?

vladucu: > Wouldn't the tests error instead of fail?

Not sure what you mean by this, but yes, an unexpected error would fail the tests. Example below with a reference error thrown in the helper
Screen Shot 2020-11-17 at 10 34 11 AM

> Isn't it better to be explicit about what we're testing, instead of saying "there aren't any assertions in this test"?

Sorry, I 💯 agree. I just can't think of anything better. Please do suggest something if you happen to know.

andrewpye: Hmm, that error message isn't helpful as it stands. Can we wrap the `click()` in a try-catch and assert if the click call blows up?

```js
try {
  await click('button');
} catch (err) {
  assert.notOk(true, 'Expected no-op not to throw when invoked');
}
```

Not a big deal if this doesn't work/make sense, just thought it'd be good to keep the tests as close to what we actually want to test as possible 🤷‍♂️

vladucu: > Hmm, that error message isn't helpful as it stands

Many errors are not helpful in JS, from my experience at least.

> Not a big deal if this doesn't work/make sense, just thought it'd be good to keep the tests as close to what we actually want to test as possible

I totally understand, no worries. I just fail to see how that's any better or closer to what we test

PR #678, opened 2020-11-12
Add support for new design language

vladucu:
Adds `polaris-app-provider` service that we use to have support for the
`newDesignLanguage`. Kind of equivalent to `polaris-react`'s `AppProvider`
component, but minimal implementation for features only for now.

**NOTE** It's just a plain service for now, likely we'll want to allow setting this by the consuming app through config to make life easier, but we can do that later


andrewpye (approved):
Neat! 🚀

PR #683, opened 2020-11-24
Update polaris banner

tomnez:
v3.10 - v5.5 diff [here](https://github.com/Shopify/polaris-react/compare/v3.10.0...v5.5.0#diff-f9882edc8324cf88b43076801e6c25a605f41b89e1e3bf19a04c7eb2c05619a7)

- [breaking] removes support for passing `@action` attribute. From now only `@primaryAction` is supported as the main action for `PolarisBanner`.
- co-locates template
- misc 🧹


addon/components/polaris-banner.hbs
```diff
@@ -1,14 +1,16 @@
 <div
   class={{this.cssClasses}}
   tabindex="0"
-  aria-live="polite"
+  role={{this.useBannerAttributes.ariaRoleType}}
+  aria-live={{if this.stopAnnouncements "off" "polite"}}
   aria-describedby={{this.contentId}}
   aria-labelledby={{this.headingId}}
-  role={{this.role}}
   {{on "mouseup" this.handleMouseUpByBlurring}}
```
vladucu, line 8: I think we need to tweak this action for `moseup` event to also set shouldShowFocus to `false` ([source](https://github.com/Shopify/polaris-react/blob/9a167d3565573739b4f2dfa8e516bc8a13a0f179/src/components/Banner/Banner.tsx#L269))

addon/components/polaris-banner.js
```diff
@@ -96,109 +79,117 @@ export default class PolarisBanner extends Component {
...
+
+  @action
+  handleBlur(e) {
+    e.currentTarget.blur();
+    this.shouldShowFocus = false;
+  }
+
+  @action
+  handleKeyUp() {
+    this.shouldShowFocus = true;
```
vladucu, line 192: not really sure if this should be a problem, but react implementation explicitly checks that the focus event is triggered on this component's main HTML element and wondering if we need to do that too _(talking about [this](https://github.com/Shopify/polaris-react/blob/9a167d3565573739b4f2dfa8e516bc8a13a0f179/src/components/Banner/Banner.tsx#L261))_

addon/components/polaris-banner.js
```diff
@@ -96,109 +79,117 @@ export default class PolarisBanner extends Component {
...
+            : 'FlagMajor',
+          iconColor: newDesignLanguage ? 'base' : 'inkLighter',
+          ariaRoleType: 'status',
+        };
+    }
+  }
+
+  @action
+  handleBlur(e) {
+    e.currentTarget.blur();
```
vladucu, line 186: I don't see this in the react implementation, is this needed?

addon/components/polaris-banner.js
```diff
@@ -96,109 +79,117 @@ export default class PolarisBanner extends Component {
...
-
-  handleMouseUpByBlurring = handleMouseUpByBlurring;
-
-  @bool('onDismiss')
-  hasDismiss;
+  @arg withinContentContainer = false;
 
-  @or('primaryAction', 'action')
-  mainAction;
+  @tracked shouldShowFocus = false;
```
vladucu, line 92: nice 🚀

vladucu (changes requested):
Looks pretty awesome to me overall. Left a few minor comments/suggestions and I think we miss something related to the `secondaryAction` which should render either an UnstyledLink of a button depending on whether the action has an `url` or not _([source](https://github.com/Shopify/polaris-react/blob/9a167d3565573739b4f2dfa8e516bc8a13a0f179/src/components/Banner/Banner.tsx#L114))_

tests/integration/components/polaris-banner-test.js
```diff
@@ -241,8 +241,34 @@ module('Integration | Component | polaris banner', function (hooks) {
...
+
+    banner = assert.dom(bannerSelector);
+
+    banner.hasAttribute(
+      'aria-live',
+      'polite',
+      'banner aria-live attribute - set to `off`'
+    );
+  });
+
```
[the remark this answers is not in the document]
tomnez: Yeah good call, I'll update 👍

addon/components/polaris-banner.js
tomnez: Ah, given that this was missing from the `handleKeyUp`, I'm guessing I got my eyes crossed and put it in here instead of the mouseup method 🤦 Yeah I'll remove.

addon/components/polaris-banner.js
tomnez: Ah, yeah not sure how needed it is but I added it anyway.

tomnez:
@vladucu good 👀, thanks for reviewing. Updates have been pushed.

addon/components/polaris-banner.hbs
```diff
@@ -0,0 +1,90 @@
...
+                    @url={{@secondaryAction.url}}
+                    @external={{@secondaryAction.external}}
+                  >
+                    <span class="Polaris-Banner__Text">{{@secondaryAction.text}}</span>
+                  </PolarisUnstyledLink>
+                {{else}}
+                  <button
+                    class="Polaris-Banner__SecondaryAction"
+                    type="button"
+                    onclick={{fn this.triggerAction @secondaryAction}}
```
vladucu, line 78:
```suggestion
                    {{on "click" (fn this.triggerAction @secondaryAction}}
```

vladucu (approved):
One minor comment, but 👍 . Thanks for the updates!
🚀

PR #689, opened 2020-12-16
Remove usage of deprecated getCode function from ember-keyboard

andrewpye:
Removes our previous usage of `getCode` from `ember-keyboard` in the `PolarisTextField` component to prevent an error being thrown when typing values into numeric text fields.

### DIFF
--- a/addon/components/polaris-text-field.js
+++ b/addon/components/polaris-text-field.js
@@ -6,7 +6,6 @@ import { htmlSafe } from '@ember/string';
 import { typeOf, isPresent } from '@ember/utils';
 import { deprecate } from '@ember/application/deprecations';
 import { tagName, layout } from '@ember-decorators/component';
-import { getCode } from 'ember-keyboard';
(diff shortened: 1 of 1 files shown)


addon/components/polaris-text-field.js
```diff
@@ -623,13 +622,13 @@ export default class PolarisTextField extends Component {
...
-    let { key } = event;
+    let { key, which } = event;
     let numbersSpec = /[\d.eE+-]$/;
 
     if (
       this.type !== 'number' ||
-      getCode(event) === 'Enter' ||
-      key.match(numbersSpec)
+      which === 13 /* Enter key */ ||
+      numbersSpec.test(key)
```
andrewpye, line 631: These changes are based on the [React implementation](https://github.com/Shopify/polaris-react/blob/master/src/components/TextField/TextField.tsx#L479).

vladucu (approved):

PR #697, opened 2021-01-12
Fix broken Glimmer component detection

andrewpye:
At some point, component definitions in production switched to using Symbol keys instead of string keys. This broke our `is-component-definition` helper, which resulted in some breakages elsewhere. This PR adds a rudimentary check for these new-style component definitions.

This is at least the second time that this helper has broken as a result of Ember/Glimmer updates. I'm aware that we should be trying to figure out a way to remove it or make it less brittle/dependent on internals, but for the time being I think this workaround will have to do 😅 

Demo of something now working after this change that was previously broken (the strong part of the help text was previously not updating because it was not receiving updates due to `is-component-definition` returning a false negative):
prod-1868-fix-component-definition-detection

### DIFF
--- a/addon/helpers/is-component-definition.js
+++ b/addon/helpers/is-component-definition.js
@@ -7,10 +7,23 @@ export function isComponentDefinition(content) {
     return false;
   }
 
+  // Glimmer now uses Symbol keys in its component definitions so we check those first.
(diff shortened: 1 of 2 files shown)


vladucu (approved):
Yeah, would be good to get rid of these hacks.

Anyway, great work here man! This will hopefully do it for now 🥇

PR #748, opened 2021-07-13
Change upgrade target to Shopify 6.6 version

vladucu:
As chatted, we're changing the target for the upgrade that's in progress from `polaris-react` `v5.5` -> `v6.6`.

- updates Shopify deps to 6.6
- adds directly a port of `geometry.js` instead of relying on `@shopify/javascript-utilities` dependency, following Shopify

### DIFF
--- a/addon/utils/geometry.js
+++ b/addon/utils/geometry.js
@@ -1,0 +1,41 @@
+class Rect {
+  static get zero() {
+    return new Rect();
+  }
+
+  constructor({ top = 0, left = 0, width = 0, height = 0 } = {}) {
+    this.top = void 0;
+    this.left = void 0;
+    this.width = void 0;
+    this.height = void 0;
+    this.top = top;
+    this.left = left;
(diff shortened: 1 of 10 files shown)


addon/utils/geometry.js
```diff
@@ -0,0 +1,41 @@
...
+    this.height = height;
+  }
+
+  get center() {
+    return {
+      x: this.left + this.width / 2,
+      y: this.top + this.height / 2,
+    };
+  }
+}
```
tomnez, line 23: nit: add a line between class/method definitions?

tomnez (approved):
Thanks @vladucu! 🙌🏻

addon/utils/geometry.js
vladucu: This is a direct copy/paste from Shopify. I've left it as is to make upgrades easier, but happy to do if you think it's worth it

PR #751, opened 2021-07-19
Enable prettier for handlebars

vladucu:


andrewpye (approved):
Good with me (not that I've checked every autofix that's been made 😅)

